PR #3, opened 2018-03-22
Ramenhog/general scaffolding

### LINKED ISSUE #2: Scaffolding the frontend
Extends this issue: https://github.com/AlexsLemonade/refinebio/issues/166

**Checklist:**
- [x] Add SASS/CSS Modules - #1 
- [x] Add redux
- [x] Add routing
- [x] Config file structure
- [x] Switch to "feature branch" model
- [x] Add PR template
- [x] Static type checking

ramenhog:
## Issue Number
Closes #2 

Laying out a general scaffold for the frontend (nothing to really see yet though)

## Types of changes

## Checklist


src/components/App/index.js
```diff
@@ -0,0 +1,28 @@
...
+import history from '../../history';
+
+import Search from '../../containers/Search';
+import Results from '../../containers/Results';
+import Experiment from '../../containers/Experiment';
+
+const App = () => {
+  return (
+    <div className={styles.App}>
+      <Router history={history}>
```
arielsvn: I hope it's not a problem, I went ahead and checked this PR. Looks good! Just one comment here

It may be a good idea to have the router in a separate file, this will make it easier if you want to change the routes later, instead of having the routes mixed with other components of the application. Like `Header` in this case.

I have used nested routes with a `Layout` component that takes care of rendering `Header` where it should be, and leaves space for its children. Something like this for example:

```jsx
const Layout = ({children}) => (
  <div className="layout">
    <div className="layout__header">
      <Header />
    </div>

    <div className="layout__content">
      {children}
    </div>
  </div>
);
```

This would allow you to have the logic concerning the layout of the app in this component, note that if you needed classes like `layout`, `layout__header` or `layout__content` to position the elements they would have to be mixed with the Routes in the current approach.

Then the Router could be something like this:

```jsx
const RouterComponent = () => (
  <Router history={history}>
    <Route path="" component={Layout}>
      <Switch>
        <Route exact path="/" component={Search} />
        <Route path="/results" component={Results} />
        <Route path="/experiment" component={Experiment} />
      </Switch>
    </Route>
  </Router>
);
```

And `App` would only have to render `<RouterComponent />`.

I'm sure there are probably other ways to achieve the same, but the important part I think is to separate the Router logic from the application. Hope this helps :)

ramenhog: Yeah, I agree, having the router in a separate component will be a good way to go in the future. But for now, a lot of these routes and components are currently just general placeholders until we get finalized designs for the app, so that'll be better suited for a future PR.

README.md
```diff
@@ -44,3 +51,41 @@ In the project directory, run:
...
 * Build is minified and filenames include hashes
+
+## Development
+
+### Git Workflow
+
+This project uses a [feature branch](http://nvie.com/posts/a-successful-git-branching-model/) based workflow.
+
+New features should be developed on new feature branches named in the following format: `[username]/[fancy-branch-name]`.
+Pull requests should be sent to the `develop` branch for code review.
```
dongbohu: The development repo in `refinebio` is named "dev", can we rename "development" branch in this repo to "dev" to make it consistent?

dongbohu:
One comment that is not really related to this PR: Line 26 of `README.md` says:
```
You can install Yarn through [Homebrew package manager](https://brew.sh/) ...
```
We probably should emphasize that `If you are using a Mac, ...`

By the way, if the user is using Linux, I suppose a `sudo npm install yarn -g` command will be good enough?

dongbohu:
You may also want to include the official URL for `Node.js` installation in `README.md`:
https://nodejs.org/en/download/
(This is how I installed Node.js and npm on Mac, because I am never a big fan of `HomeBrew`.)

ramenhog:
@dongbohu Added Node and Yarn installation links to the readme and also changed the default branch to `dev` to avoid future confusion. 👍 Back to you, thanks!

ramenhog:
👍 deleted

PR #4, opened 2018-04-12
Ramenhog/exec dashboard

ramenhog:
## Issue Number
https://github.com/AlexsLemonade/refinebio/issues/177

Create an executive dashboard to view stats/charts about jobs/experiments/samples etc.

## Types of changes

## Screenshots
screen shot 2018-04-12 at 5 13 44 pm
screen shot 2018-04-12 at 5 13 59 pm


src/state/dashboard/reducer.js
```diff
@@ -0,0 +1,179 @@
...
+      return state;
+    }
+  }
+};
+
+export default dashboardReducer;
+
+// chart selectors for creating chart data for individual charts on dashboard
+
+export function getTotalLengthofQueuesByType(state) {
```
dongbohu: "...LengthOf..." (up to you)

src/state/dashboard/actions.js
```diff
@@ -0,0 +1,180 @@
...
+          .utc()
+          .format();
+
+    const response = await fetch(
+      `${endPoint}?created_at__gte=${gte}&created_at__lte=${lte}&limit=${limit}`
+    );
+
+    if (response.status === 200) {
+      return await response.json();
+    } else {
```
dongbohu: I guess you can delete lines 52 and 58 to make the code a little more compact (up to you).

src/state/dashboard/actions.js
```diff
@@ -0,0 +1,180 @@
...
+          error: 'there was an error with response'
+        }
+      ];
+    }
+  });
+  return Promise.all(promiseArray);
+}
+
+function getJobStatusesOverTime(jobData = []) {
+  const pending = jobData.map(
```
dongbohu: So `pending`, `failed`, `completed` and `open` are all arrays that indicate each job's status? Wouldn't it be better to scan `jobData` once, check each job's `start_time` and/or `success` fields to determine its status? You only need one array as a return value. 

How many elements does `jobData` has typically? If it is not that big, then 1 scan or 4 scans doesn't really matter that much.

src/state/dashboard/reducer.js
```diff
@@ -0,0 +1,179 @@
...
+
+export function getAllEstimatedTimeTilCompletion(state, jobType) {
+  const stats = state.dashboard.stats;
+
+  if (!Object.keys(stats).length) return {};
+
+  return Object.keys(stats).reduce((allEstimatedTimes, jobType) => {
+    const estimateSec =
+      (stats[jobType].open + stats[jobType].pending) *
+      parseInt(stats[jobType].average_time, 10);
```
dongbohu: Wouldn't `Math.round()` make more sense here?

src/state/dashboard/reducer.js
```diff
@@ -0,0 +1,179 @@
...
+      date: time.utc().format()
+    };
+
+    Object.keys(jobs).forEach(jobName => {
+      if (jobs[jobName].all.length === timePoints.length) {
+        // accumulate count completed over time
+        const jobCount =
+          i === 0
+            ? jobs[jobName].all[i].count
+            : jobTotals[i - 1][jobName] + jobs[jobName].all[i].count;
```
dongbohu: Can you explain to me what this addition means? What about the `jobName` field of the last element in `jobTotals`?

Miserlou:
Needs documentation and live-updating for this to be useful as a dashboard that we can leave on. It'd be great if the default view was hour-wise or ever-15-minute-wise - daily is too broad for us to make good real time usage of it. Charts look great!

src/index.scss
```diff
@@ -1,5 +1,68 @@
...
+.loader {
+  width: 100px;
+  height: 100px;
+  border-radius: 100%;
+  position: relative;
+  margin: 0 auto;
+  -webkit-perspective: 120px;
+  -moz-perspective: 120px;
+  -ms-perspective: 120px;
+  perspective: 120px;
```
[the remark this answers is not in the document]
ramenhog: Apparently sometimes the error highlighting is wrong: https://stackoverflow.com/a/44631734/9383002

src/components/PieChart/index.js
```diff
@@ -0,0 +1,43 @@
...
+          data={data}
+          dataKey="value"
+          nameKey="name"
+          cx="50%"
+          cy="50%"
+          outerRadius={'100%'}
+          fill="#8884d8"
+          paddingAngle={0}
+        >
+          {data.map((entry, index) => (
```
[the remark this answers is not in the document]
ramenhog: It is not. This is just to create the different colors for each cell.

src/components/TimeRangeSelect/TimeRangeSelect.scss
```diff
@@ -0,0 +1,57 @@
+@import '../../common/styles/variables';
+
+.time-range-select {
+  justify-content: flex-end;
+  display: flex;
+
+  &__label {
+    margin-right: 16px;
```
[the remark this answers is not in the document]
ramenhog: Deepa had introduced me to designing with an 8-point grid, so I've just been using multiples of 8 in my spacing :)

src/components/LineChart/index.js
```diff
@@ -0,0 +1,56 @@
...
+      <LineRechart
+        data={data}
+        margin={{ top: 5, right: 30, left: 0, bottom: 5 }}
+      >
+        <XAxis dataKey="date" tickFormatter={formatXAxis} />
+        <YAxis />
+        <CartesianGrid strokeDasharray="3 3" />
+        <Tooltip />
+        <Legend />
+        {series.map((set, i) => (
```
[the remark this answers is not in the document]
ramenhog: It might not be...I mainly chose that word because there's no singular word for "series" haha

src/state/dashboard/reducer.js
ramenhog: `average_time` seems to get returned as a string, so I need to make it an integer.

dongbohu: What I meant was that using `Math.round()` will be a little more accurate. But since this is just an estimate, it doesn't really matter that much whether a value like `9.8` is converted to 9 or 10.

ramenhog: ahh yes, probably `parseFloat()` would actually work best here, so it doesn't remove the decimals

src/state/dashboard/reducer.js
ramenhog: So, if this is the first date in the time range, `jobCount` will just be equal to the `count` for that date. If not, I accumulate the total of jobs run from the first date in the time range up to this current date for `jobCount`.

src/state/dashboard/actions.js
dongbohu: Suppose `jobData` is a list like this:
```
[ job_group_1, job_group[_2, job_group_3, ... ]
```
You can count the the number of pending/failed/open/complete/ by scanning `jobData` only once, 
instead of scan once to count all jobs that are `pending`, 2nd scan to count `failed` jobs, 3rd scan for `completed`, 4th scan for `open`. (Each call of `map` and `filter` is a scan of the whole list.)

As I mentioned earlier, if the list is short, there is no big deal. But if the list has thousands (or even more) jobs, it may make some difference.

Does this make sense?

src/containers/Dashboard/index.js
```diff
@@ -15,8 +15,16 @@ class Dashboard extends Component {
       updatedTimeRange,
       fetchDashboardData
     } = this.props;
-    updatedTimeRange(range);
-    fetchDashboardData();
+
+    // data live reloads every 5 seconds
+    setInterval(
+      (function fetchData() {
```
dongbohu: Isn't it same as:
```
setInternal(
  function fetchData() {
    updatedTimeRange(range);
    fetchDashboardData();
  },
  5000
);
```

ramenhog: I made it a self-invoking function because I want `fetchData` to run immediately and then be called every 5 seconds.

src/state/dashboard/reducer.js
```diff
@@ -0,0 +1,196 @@
...
+      },
+      []
+    );
+    return accum;
+  }, {});
+}
+
+function convertSecToMinHours(sec) {
+  const hours = Math.floor(sec / 3600),
+    minutes = Math.floor((sec % 3600) / 60);
```
dongbohu: Since we only estimate the time to the minute level, I think it will be slightly better to use `Math.round()` here.

src/state/dashboard/reducer.js
```diff
@@ -87,8 +87,8 @@ export function getJobsByStatus(state) {
 }
 
 function convertSecToMinHours(sec) {
-  const hours = Math.floor(sec / 3600),
-    minutes = Math.floor((sec % 3600) / 60);
+  const hours = Math.round(sec / 3600),
```
dongbohu: For hours, you actually don't want to round it because we only care about the integer part.

dongbohu: Only the value in the finest unit should be rounded.

src/containers/Dashboard/index.js
```diff
@@ -0,0 +1,233 @@
...
+    } = this.props;
+
+    fetchDashboardData();
+
+    // data live reloads every 5 seconds
+    this.intervalId = setInterval(
+      (function fetchData() {
+        updatedTimeRange(range);
+        return fetchData;
+      })(),
```
dongbohu: This idea of invoking the callback function immediately is interesting. I never thought it could be done like that. Is it a common pattern in JS?

ramenhog: Hmm, I just know I've seen it used before for this purpose, so I've begun using it myself :)

PR #10, opened 2018-05-17
Add Shareable Link Modal to Download page

arielsvn:
## Issue Number

#9 

- Added a generic component for dialogs
- Added Shareable Link Modal to Download page 

We will still need to connect Redux to show the correct url.

## Types of changes

## Screenshots

download

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -1,14 +1,14 @@
 import React from 'react';
 import './Button.scss';
 
-const Button = ({ text, buttonStyle, onClick, className }) => {
+const Button = ({ text, buttonStyle, onClick, className, children }) => {
   const buttonStyleClass = buttonStyle ? `button--${buttonStyle}` : '';
   return (
     <button
-      className={`button ${buttonStyleClass} ${className}`}
+      className={`button ${buttonStyleClass} ${className || ''}`}
       onClick={onClick}
     >
-      {text}
+      {text || children}
     </button>
   );
 };
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -1,10 +1,13 @@
 import React from 'react';
 import './Button.scss';
 
-const Button = ({ text, buttonStyle, onClick }) => {
+const Button = ({ text, buttonStyle, onClick, className }) => {
   const buttonStyleClass = buttonStyle ? `button--${buttonStyle}` : '';
   return (
-    <button className={`button ${buttonStyleClass}`} onClick={onClick}>
+    <button
+      className={`button ${buttonStyleClass} ${className}`}
+      onClick={onClick}
+    >
       {text}
     </button>
   );
--- a/src/components/Input/Input.scss
+++ b/src/components/Input/Input.scss
@@ -1,10 +1,10 @@
-
+@import '../../common/styles/variables';
 
 .input {
     font-size: 16px;
     line-height: 24px;
     padding: 4px 8px;
-    border: 1px solid #9B9B9B;
+    border: 1px solid $input-normal;
     border-radius: 1px;
     outline: none;
     width: 100%;
@@ -14,7 +14,7 @@
     }
 
     &:focus {
-        border-color: #386DB0;
+        border-color: $input-focus;
     }
 }
 
--- a/src/components/Input/Input.scss
+++ b/src/components/Input/Input.scss
@@ -1,0 +1,22 @@
+
+
+.input {
+    font-size: 16px;
+    line-height: 24px;
+    padding: 4px 8px;
+    border: 1px solid #9B9B9B;
+    border-radius: 1px;
+    outline: none;
+    width: 100%;
+    
+    &:hover {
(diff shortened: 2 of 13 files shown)


src/components/InputCopy/index.js
```diff
@@ -0,0 +1,52 @@
...
+      <div className="input-copy">
+        {this.state.alert && (
+          <div className="input-copy__alert">
+            <i className="ion-checkmark-circled" /> Copied to clipboard.
+          </div>
+        )}
+        <div>
+          <Button
+            text="Copy"
+            className="input-copy__btn"
```
ramenhog: From the screenshot it looks like a standard primary button, so you shouldn't have to pass in a custom `className`?

src/components/Input/Input.scss
```diff
@@ -0,0 +1,22 @@
+
+
+.input {
+    font-size: 16px;
+    line-height: 24px;
+    padding: 4px 8px;
+    border: 1px solid #9B9B9B;
```
ramenhog: Let's replace all the hexcodes with sass variables in the `common/variables` file for consistency

src/components/Input/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import './Input.scss';
+
+type Props = {
```
ramenhog: Nice! Type checking 🎉

src/components/Button/index.js
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
 import './Button.scss';
 
-const Button = ({ text, buttonStyle, onClick }) => {
+const Button = ({ text, buttonStyle, onClick, className }) => {
   const buttonStyleClass = buttonStyle ? `button--${buttonStyle}` : '';
   return (
-    <button className={`button ${buttonStyleClass}`} onClick={onClick}>
+    <button
+      className={`button ${buttonStyleClass} ${className}`}
```
ramenhog: We should probably check to make sure className isn't `null` or `undefined`

src/components/InputCopy/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+export default class InputCopy extends React.Component {
+  state = {
+    alert: false
+  };
+
+  render() {
+    return (
+      <div className="input-copy">
+        {this.state.alert && (
```
ramenhog: 👍

src/components/Modal/index.js
```diff
@@ -0,0 +1,90 @@
...
+          className={`modal ${
+            this.props.center ? 'modal--center' : 'modal--top'
+          }`}
+          // Check that clicks were originated exactly on the `.modal` element, aka outside of the
+          // modal dialog. Clicks inside of the modal dialog will trigger this handler too.
+          onClick={e => e.target === e.currentTarget && this._closeModal()}
+        >
+          <div className={`modal__dialog ${this.props.className}`}>
+            {this.props.children}
+            <a
```
ramenhog: For accessibility purposes, let's actually change this to a button.

src/components/InputCopy/index.js
arielsvn: hmm that class is used to position the button to the right. Do you think it's better to add an additional element containing the button and apply the class to that one instead?

ramenhog: hmm what about just doing `display: flex` on the parent?

arielsvn: haha great idea! I have to start using `flex` more

arielsvn:
@ramenhog I addressed all the comments, let me know if you see something else

src/components/Modal/ModalManager.js
```diff
@@ -0,0 +1,34 @@
...
+    modalOpen: false
+  };
+
+  render() {
+    return (
+      <React.Fragment>
+        {this.props.component(() => this.setState({ modalOpen: true }))}
+        <Modal
+          {...this.props.modalProps}
+          key="2"
```
ramenhog: are we actually using this `key` anywhere?

src/components/Modal/index.js
```diff
@@ -0,0 +1,91 @@
...
+    this.el = document.createElement('div');
+  }
+
+  componentDidMount() {
+    modalRoot.appendChild(this.el);
+    this._addBodyClass();
+  }
+
+  componentDidUpdate() {
+    this._addBodyClass();
```
ramenhog: I'm wondering if we need a conditional check here to update the body class only when `isOpen` changes?

src/components/Modal/ModalManager.js
arielsvn: No, we're not using it anymore, I forgot to remove it after refactoring that. Let me do it.

src/components/Modal/index.js
arielsvn: good idea, that way we can avoid adding/removing the classes from the body unnecessarily.

arielsvn:
yes me too, I like what we did there :)

PR #35, opened 2018-05-24
Link to Samples table from search results

### LINKED ISSUE #17: Search Page: Link "View Samples" button to the samples section on the experiment page
### Description
When user clicks on "View Samples" button for an individual experiment in their search results, it should take them to the experiment page for that experiment and anchor them at the samples table section.

arielsvn:
## Issue Number

closes #17 

When user clicks on "View Samples" button for an individual experiment in their search results, it should take them to the experiment page for that experiment and anchor them at the samples table section.

## Types of changes

## Functional tests

- In Results page, clicked View Samples, and it displayed the samples table directly

## Screenshots

anchor2

### DIFF
--- a/src/components/Anchor/index.js
+++ b/src/components/Anchor/index.js
@@ -25,6 +25,6 @@ export default class Anchor extends React.Component {
   }
 
   render() {
-    return <a ref={x => (this.element = x)} name={this.props.name} />;
+    return <div ref={x => (this.element = x)}>{this.props.children()}</div>;
   }
 }
--- a/src/components/Anchor/index.js
+++ b/src/components/Anchor/index.js
@@ -1,12 +1,6 @@
 import React from 'react';
 
 /**
- * This constant is used to correct the scroll bar position when navigating to an anchor
- * Since we want the target element to appear below the scrollbar
- */
-const TOP_BAR_HEIGHT = 110;
-
-/**
  * Creates an anchor class that scrolls to its possition if the correct hash has been specified
  * Only when the component is loaded
  */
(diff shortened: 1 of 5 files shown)


src/components/Button/Button.scss
```diff
@@ -38,3 +38,11 @@
     color: $black;    
   }
 }
+
+
+// For some reason `a` elements styled as buttons don't have the same height.
+// This rule fixes that
+a.button {
```
ramenhog: This is happening because `a` are inline elements so you can't add height to the top or bottom padding/margin. We can fix this with `inline-block` on the `.button` class. 🚫 tag selectors

src/components/Anchor/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+/**
+ * This constant is used to correct the scroll bar position when navigating to an anchor
+ * Since we want the target element to appear below the scrollbar
+ */
+const TOP_BAR_HEIGHT = 110;
```
ramenhog: Can we get this with JS? That way, if we change the height of the navbar in the future, we don't have to manually update it here.

src/components/Button/Button.scss
arielsvn: Interesting! let me give it a try

arielsvn:
> Is there any way to just add an id to the h2 and link to that from the search results? Just wondering

hmm maybe, I wanted this to be an independent component that we can use with `<Anchor name="sample" />`... what if we change the `<a>` that is currently being rendered for a `<span>`?  That will remove the warning

arielsvn:
Using a render prop component for the `Anchor` was a great idea! Thanks @ramenhog! I also addressed the other comments, let me know what do you think.

src/components/Anchor/index.js
```diff
@@ -19,12 +13,24 @@ export default class Anchor extends React.Component {
...
+  /**
+   * To work properly this component needs to measure the height of the header, to be able to calculate
+   * the correct position of the scrollbar when navigating to an element in the screen.
+   *
+   * This method is probably an anti-pattern in React, since it creates a dependency between this component and
+   * the header component that is in other place. A better solution could take advantage of the React Context API
+   * but that seemed overcomplicated for this use case.
+   */
+  _topBarHeight() {
+    return document.getElementsByClassName('js-header')[0].offsetHeight;
```
ramenhog: 🙌

PR #37, opened 2018-05-25
Ramenhog/feature/dynamic dataset summary data

### LINKED ISSUE #24: Downloads Page: Populate Dataset Summary with endpoint data
## Description
The Dataset Summary section is currently hardcoded. Now that we get species information via the experiment endpoint, we should dynamically populate the table

## Mockup
screen shot 2018-05-18 at 2 44 15 pm

ramenhog:
## Issue Number
Closes #24 

The "DataSet Summary" section now dynamically generates the stats via the user's dataset 

## Types of changes

New feature (non-breaking change which adds functionality)

## Functional tests
* stats update as user adds/removes from their dataset
* the species rows are removed when there is an empty dataset

## Screenshots

issue-24

### DIFF
--- a/src/containers/Downloads/DownloadDatasetSummary.js
+++ b/src/containers/Downloads/DownloadDatasetSummary.js
@@ -1,6 +1,24 @@
 import React from 'react';
 
-const DownloadDatasetSummary = () => {
+const DownloadDatasetSummary = ({ species }) => {
+  const renderSpeciesRow = species => {
+    console.log(species);
+    return Object.keys(species).map(organism => {
+      const samples = species[organism];
+      return (
+        <tr className="downloads__table-row">
+          <td className="downloads__table-cell">{organism}</td>
+          <td className="downloads__table-cell downloads__table-cell--value">
+            {samples.length}
+          </td>
+          <td className="downloads__table-cell downloads__table-cell--value">
+            10
+          </td>
+        </tr>
+      );
+    });
+  };
+
   return (
     <section className="downloads__section">
       <h2>Dataset Summary</h2>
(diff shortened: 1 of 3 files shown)


src/containers/Downloads/index.js
```diff
@@ -178,7 +194,12 @@ class Download extends Component {
         ) : (
           <div>
             <DownloadFileSummary summaryData={downloadFilesData} />
-            <DownloadDatasetSummary data={this.props.dataSet} />
+            <DownloadDatasetSummary
```
arielsvn: Is it possible to use connect on `DownloadDatasetSummary` and send these properties directly into it? (`samplesBySpecies`, `totalSamples`, `totalExperiments` and `experimentCountBySpecies`) Instead of putting them on `Download` and passing them down

src/containers/Downloads/DownloadDatasetSummary.js
```diff
@@ -14,33 +50,8 @@ const DownloadDatasetSummary = () => {
...
-              <td className="downloads__table-cell">Total</td>
-              <td className="downloads__table-cell downloads__table-cell--value">
-                100
-              </td>
-              <td className="downloads__table-cell downloads__table-cell--value">
-                10
-              </td>
-            </tr>
+            {renderSpeciesRow()}
+            {renderTotalsRow()}
```
arielsvn: I like this refactor! Maybe it would be good to extract these functions into separate components, or use the class notation for `DownloadDatasetSummary`. Otherwise `renderSpeciesRow` and `renderTotalsRow` will be redefined on every render.

src/containers/Downloads/index.js
ramenhog: Hmm it is possible, but I'm not sure it's necessary to connect it since the parent needs and has access to those props already.

arielsvn: That's true, I didn't realize of those usages.

PR #44, opened 2018-05-29
Arielsvn/feature/32 download link

### LINKED ISSUE #32: Downloads Page: Share Download Link
## Description
* when a user clicks the "Share" button on their download, they should be given a url with either `/download/:dataSetId` or `/download?view=:dataSetId` to share
* when users load that shared download link with params, they should be shown the details of the dataset associated with that ID but do **not** have the ability remove samples from the dataset
   * they can only "Share" and "Download"

arielsvn:
## Issue Number

closes #32 

Add additional Download page, so that users can view a read only version when someone else shares a download url.

## Types of changes

## Functional tests

- Open downloads page
- Open downloads page with shared link http://localhost:3000/download/2bf1c9a5-ce16-4d4a-925e-74974509b025, here it doesn't show buttons to edit download
- Share download modal has the correct id on both cases.

## Checklist

## Screenshots

image

### DIFF
--- a/src/components/TabControl/index.js
+++ b/src/components/TabControl/index.js
@@ -1,6 +1,11 @@
 import React from 'react';
 import Toggle from '../Toggle';
 
+/**
+ * Renders a simple tab control with a `Toggle` at the top, each tab item can be specified
+ * as children of this component. At any given time only one of those childrens will be rendered
+ * The one that has the same index as the selected tab.
+ */
 export default class TabControl extends React.Component {
   state = {
     tabIndex: 0
--- a/src/components/TabControl/index.js
+++ b/src/components/TabControl/index.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import Toggle from '../Toggle';
+
+export default class TabControl extends React.Component {
+  state = {
+    tabIndex: 0
+  };
+
+  handleTabChange = tabIndex => {
+    this.setState({
+      tabIndex
+    });
+  };
+
+  render() {
+    return (
+      <div>
+        <Toggle tabs={this.props.tabs} onToggle={this.handleTabChange} />
+        {React.Children.toArray(this.props.children)[this.state.tabIndex]}
+      </div>
+    );
+  }
+}
--- a/src/containers/Downloads/DownloadBar.js
+++ b/src/containers/Downloads/DownloadBar.js
@@ -4,7 +4,6 @@ import Dropdown from '../../components/Dropdown';
 import ModalManager from '../../components/Modal/ModalManager';
 import InputCopy from '../../components/InputCopy';
 import './DownloadBar.scss';
-import { connect } from 'react-redux';
 
 let DownloadBar = ({ dataSetId, shareUrl }) => {
   return (
--- a/src/containers/Downloads/DownloadBar.js
+++ b/src/containers/Downloads/DownloadBar.js
@@ -6,7 +6,7 @@ import InputCopy from '../../components/InputCopy';
 import './DownloadBar.scss';
 import { connect } from 'react-redux';
 
-let DownloadBar = ({ shareUrl }) => {
+let DownloadBar = ({ dataSetId, shareUrl }) => {
   return (
     <div className="downloads__bar">
       <ModalManager
@@ -18,7 +18,9 @@ let DownloadBar = ({ shareUrl }) => {
         {() => (
           <div>
             <h1 className="share-link-modal__title">Sharable Link</h1>
-            <InputCopy value={shareUrl} />
+            <InputCopy
+              value={`${window.location.origin}/download/${dataSetId}`}
+            />
           </div>
         )}
       </ModalManager>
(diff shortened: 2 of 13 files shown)


src/containers/Downloads/DownloadBar.js
```diff
@@ -19,7 +19,9 @@ let DownloadBar = ({ startDownload }) => {
         {() => (
           <div>
             <h1 className="share-link-modal__title">Sharable Link</h1>
-            <InputCopy value="Url to be copied, connect redux" />
+            <InputCopy
+              value={`${window.location.origin}/download/${dataSetId}`}
```
ramenhog: Actually, it might be nice to store the domain in a constants file. And we could have one for `DEV` and one for `PROD`.

src/components/TabControl/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Toggle from '../Toggle';
+
+/**
+ * Renders a simple tab control with a `Toggle` at the top, each tab item can be specified
+ * as children of this component. At any given time only one of those childrens will be rendered
+ * The one that has the same index as the selected tab.
+ */
+export default class TabControl extends React.Component {
```
ramenhog: Nice refactor 👍

src/state/download/actions.js
```diff
@@ -174,58 +175,34 @@ export const addExperimentSucceeded = (dataSetId, dataSet) => {
 export const fetchDataSet = () => {
   return async dispatch => {
     const dataSetId = localStorage.getItem('dataSetId');
+    if (!dataSetId) {
+      return dispatch(fetchDataSetSucceeded({}));
```
ramenhog: Why are we calling this success function when there's no `dataSetId` found?

src/state/viewDownload/reducer.js
```diff
@@ -0,0 +1,16 @@
+const initialSatate = {};
```
ramenhog: ahh, a typo 🙃

src/state/viewDownload/reducer.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default function(state = initialSatate, action) {
+  const { type, data } = action;
+
+  switch (type) {
+    case 'LOAD_VIEW_DOWNLOAD': {
+      return {
+        ...state,
+        ...data
```
ramenhog: This is more a personal opinion, but might be nice to actually explicitly declare all the keys that are going to be part of state instead of spreading so we can have a clearer representation of what is being put in this slice of state.

src/state/viewDownload/actions.js
```diff
@@ -0,0 +1,17 @@
+import { getDataSet, getSamplesAndExperiments } from '../download/dataSetApi';
```
ramenhog: This is great!

arielsvn: Glad you like it :) Here I wasn't sure where to put the file `dataSetApi`, I placed inside the `download` reducer folder, but it's also used from the `viewDownload` reducer.

src/state/viewDownload/reducer.js
arielsvn: I agree, good idea!

src/state/viewDownload/actions.js
ramenhog: Hmm, maybe we can have a specific `api` folder? I could see that being useful for other api calls in the future.

src/state/download/actions.js
arielsvn: hmm yes, I refactored this code to remove one condition, before it was:

```
    const dataSetId = localStorage.getItem('dataSetId');
    [...]
    const response = dataSetId
      ? await asyncFetch(`/dataset/${dataSetId}/`)
      : null;

    const dataSet = response ? response.data : {};

    dispatch(fetchDataSetSucceeded(dataSet));
```

So even when `dataSetId` is not defined it still calls the success function with an empty object. Should we remove this part??

src/state/viewDownload/actions.js
arielsvn: ok let's do that 👍

src/state/download/actions.js
ramenhog: haha my bad. Yes, my gut instinct is that we should only call the success function when there is a dataSetId? What do you think?

src/containers/Downloads/DownloadBar.js
arielsvn: hmm I'm not sure about this, because it will increase the complexity to deploy the project. 

We'll need one value for `STAGING`, and if someone runs the project locally in a different domain than `localhost:3000` they it will need to update the configuration file. What advantages you see in using a constant over this approach??

ramenhog: Yeah, I see your point. We can stick with the `window.location` approach 👍

arielsvn: what if we extract that code into a helper? `helpers.getDomain` or something like that. That way it will be easier if in the future we introduce a constant it will be easier to change

ramenhog: Good call. Yeah, that would probably be helpful because I think we might have to introduce environment variables eventually anyways because the api domains will need to be based on the environment the app is in.

arielsvn:
Thanks @ramenhog! :) I pushed some fixes, let me know what do you think.

PR #49, opened 2018-05-31
Add Submitter processed modal to Samples Table

arielsvn:
## Issue Number

#22

## Purpose/Implementation Notes

Depends on https://github.com/AlexsLemonade/refinebio-frontend/pull/48

Adds the Submitter Processed dialog, when clicking on Processing Information in the Samples table.

## Types of changes

* [x] New feature (non-breaking change which adds functionality)

## Functional tests

- Confirmed that the Processing Information of Samples processed by the submitter is *Submitter-processed*
- When clicking there, the [modal dialog](https://projects.invisionapp.com/d/main#/console/13557531/297508606/preview) is displayed

## Screenshots

image

PR #50, opened 2018-06-01
Add Processing Information Modal Dialogs

### LINKED ISSUE #22: Experiment Page: Add processing information modals
## Description
When user clicks on the link in the "Processing Information" column in the samples table, they should get a modal that displays all the processing information associated with that sample.

### refine.bio processed with 1 middle step
screen shot 2018-05-18 at 2 25 22 pm
### refine.bio processed with 2 middle steps
screen shot 2018-05-18 at 2 26 43 pm
### submitter processed
screen shot 2018-05-18 at 2 26 23 pm

arielsvn:
## Issue Number

#22 

https://github.com/AlexsLemonade/refinebio-frontend/pull/49 Adds one processing dialog and this PR adds the remaining two.

## Types of changes

## Functional tests

- Checked that both dialogs were displayed correctly 

## Screenshots

image

image


arielsvn:
@ramenhog, in case you want to take another look, I added a new commit with the logic to decide which modal dialog to display based on the `pipelines` field https://github.com/AlexsLemonade/refinebio-frontend/pull/50/commits/010869b76b1d0cf207b7821fbd53d10fc1025f0a

This PR depends on https://github.com/AlexsLemonade/refinebio/pull/299.

src/containers/Experiment/SamplesTable.js
```diff
@@ -284,19 +284,58 @@ function ThComponent({ toggleSort, className, children, ...rest }) {
...
+  ) {
+    // check for `pipelines === ['tximport', 'Salmon']`
+    return <SalmonTxtimportModal sample={sample} />;
+  } else if (
+    pipelines.length === 2 &&
+    pipelines[0] === PIPELINES.Salmon &&
+    pipelines[1] === PIPELINES.tximport
+  ) {
+    // check for `pipelines === ['Salmon', 'tximport']`
+    return <SalmonTxtimportModal sample={sample} />;
```
ramenhog: Correct me if I'm wrong, but the checks for lines 319-325 and 326-332 seem to result in the same thing and the only thing different in the conditionals is the order of the pipelines? Am I understanding that correctly?

arielsvn: Yes that's correct. I added it like that trying to make an emphasis that we're testing those two values specifically pipelines === ['tximport', 'Salmon'] or pipelines === ['Salmon', 'tximport'].

Do you think we should put these conditions together??

ramenhog: Yeah, in that case, I think we might be able to simplify the checks in this function. Maybe something like this:

```
if (pipelines.length) {
   	if (pipelines.includes(PIPELINE_NAME) {
		return <AssociatedComponent />;

Comment (arielsvn) [reply src/containers/Experiment/SamplesTable.js]: hmm but with `includes` we wouldn't be checking for exact values of the `pipelines` array... and we want to ensure it only works for the combinations we defined. What do you think of this approach ??

```
  const pipelineString = pipelines.join(', ');
  
  const pipelineFactory = {
    'Affymetrix SCAN': () => <ScanModal sample={sample} scanType={PIPELINES.AffymetrixSCAN} />,
    'Illumina SCAN': () => <ScanModal sample={sample} scanType={PIPELINES.IlluminaSCAN} />,
    'Submitter-processed': () => <SumitterProcessedModal sample={sample} />,
    'tximport, Salmon': () => <SalmonTxtimportModal sample={sample} />,
    'Salmon, tximport': () => <SalmonTxtimportModal sample={sample} />,
  };

  if (pipelineFactory[pipelineString]) {
    return pipelineFactory[pipelineString]();
  } else {
    return <div>{pipelineString}</div>;
  }

```

PR #52, opened 2018-06-01
Arielsvn/support/23 samples table sorting

### LINKED ISSUE #23: Experiment Page: display correct fields in samples table
## Description
Display the fields we want as columns in the samples table.

**Relies on:** https://github.com/AlexsLemonade/refinebio/pull/252

arielsvn:
## Issue Number

Close #23 

Used sample details endpoint to fetch the samples associated with an experiment. Also added the ability to sort the table passing the corresponding parameters to the endpoint.

The request urls will looks like: http://localhost:3000/samples/?ids=241,242,243,244,245,246,247,248,249,250,251,252,253,254,255,256,257,258,259,260,261,262,263,264,265,266,267,268,269,270,271,272,273,274,275,276,277,278,279,280,281,282,283,284,285,286,287,288,289,290,291,292,293,294,295,296,297,298,299,300,301,302,303,304&offset=0&limit=10&order_by=-title

## Types of changes

## Functional tests

- Check that the table renders the samples associated with an experiment
- Sort table and check pagination

## Screenshots

samples sorting

### DIFF
--- a/src/api/samples.js
+++ b/src/api/samples.js
@@ -1,4 +1,4 @@
-import { asyncFetch } from '../common/helpers';
+import { asyncFetch, Ajax } from '../common/helpers';
 
 /**
  * Returns detailed information for the given sample id
@@ -10,8 +10,17 @@ export async function getDetailedSample(sampleId) {
 
 /**
  * Returns all the details of the samples ids given
- * @param {Array} sampleIds Ids of the samples
+ * @param {Array} ids Ids of the samples
+ * @param {orderBy} ids Field to sort the samples eg(title or -title). ref https://github.com/AlexsLemonade/refinebio/pull/298#issue-191878045
+ * @param {number} offset
+ * @param {number} limit
  */
-export async function getAllDetailedSamples(sampleIds) {
-  return Promise.all(sampleIds.map(id => getDetailedSample(id)));
+export async function getAllDetailedSamples({ ids, orderBy, offset, limit }) {
+  let { results } = await Ajax.get('/samples/', {
+    ids,
+    offset,
+    limit,
+    order_by: orderBy
+  });
+  return results;
 }
--- a/src/common/helpers.js
+++ b/src/common/helpers.js
@@ -68,3 +68,24 @@ export function getAmazonDownloadLinkUrl(s3_bucket, s3_key) {
 export function getDomain() {
   return window.location.origin;
 }
+
+// Helper methods to ease working with ajax functions
+export const Ajax = {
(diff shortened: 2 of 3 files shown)


src/common/helpers.js
```diff
@@ -68,3 +68,24 @@ export function getAmazonDownloadLinkUrl(s3_bucket, s3_key) {
 export function getDomain() {
   return window.location.origin;
 }
+
+// Helper methods to ease working with ajax functions
+export const Ajax = {
```
ramenhog: This is great! 👏 Can you refactor all the instances where we're doing requests?

src/api/samples.js
```diff
@@ -10,8 +10,17 @@ export async function getDetailedSample(sampleId) {
...
  * Returns all the details of the samples ids given
- * @param {Array} sampleIds Ids of the samples
+ * @param {Array} ids Ids of the samples
+ * @param {orderBy} ids Field to sort the samples eg(title or -title). ref https://github.com/AlexsLemonade/refinebio/pull/298#issue-191878045
+ * @param {number} offset
+ * @param {number} limit
  */
-export async function getAllDetailedSamples(sampleIds) {
-  return Promise.all(sampleIds.map(id => getDetailedSample(id)));
+export async function getAllDetailedSamples({ ids, orderBy, offset, limit }) {
```
ramenhog: I'm curious why you choose to pass in the whole thing as an object instead of individual arguments?

src/common/helpers.js
arielsvn: Sure! but do you think we should include that in this PR? It may be good to review this refactor separately to ensure we don't introduce any bugs.

src/api/samples.js
arielsvn: haha I commented the same here https://github.com/AlexsLemonade/refinebio-frontend/pull/51#discussion_r192488331, mostly personal preference I think. When there're more than 3-4 parameters, an object makes it easier to manage them since we don't have to worry about the order they have

arielsvn:
> when you sort the samples but you're not on the first page, does it show the current page for sorted results? or does it just sort the subset of samples from that page?

It shows the current page but sorting with the new field. So if you're in page 3, and change the column that is sorting the samples, it will show page 3 of the elements sorted by the new column. All the sorting happens in the backend, we ask for the same page but changing the `order_by` parameter.

src/common/helpers.js
ramenhog: yeah, that works for me 👍

src/api/samples.js
ramenhog: I like that reasoning 👍

Issue #53, opened 2018-06-04
Bug on ModalManager


arielsvn:
It seems the backdrop doesn't fill the entire space in some cases, and it ads several `div` elements to the body on the Downloads page.

arielsvn:
I identified what was causing this bug: we have to add a class to the body element (`modal-open`), but since we have several independent instances of `Modal` they were having conflicts with the others.

I tried to look for better options to implement the modals and found [this answer](https://stackoverflow.com/a/35641680/763705) from Dan Abramov with some tips on how to do it with Redux. There he writes:

> Instead of giving you advice on implementing this, I suggest that you just don’t implement it yourself. It is hard to get right considering accessibility.
> 
> Instead, I would suggest you to use an accessible off-the-shelf modal component such as react-modal. [..]

@ramenhog how do you feel about using that package https://github.com/reactjs/react-modal ??

It seems to be well maintained and it provides a component similar to the `Modal` we have.

PR #54, opened 2018-06-04
Connect view samples table on downloads page

### LINKED ISSUE #26: Downloads Page: Connect "View Samples" table
## Description
When user clicks "View Samples" button, they should see a table listing all the samples associated with the experiment/species.

arielsvn:
## Issue Number

Close #26 

Adds modals dialogs on the Download page with the samples associated to each experiment and specie.

I noticed a bug on the ModalManager, and reported it on #53

## Types of changes

## Functional tests

- Open downloads page, and click view samples on the species
- Same for experiments on the other tab

## Screenshots

Samples for Homo Sapiens specie:

image

Samples for experiment:

image

### DIFF
--- a/src/containers/Experiment/SamplesTable.js
+++ b/src/containers/Experiment/SamplesTable.js
@@ -26,7 +26,7 @@ export default class SamplesTable extends React.Component {
   }
 
   render() {
-    const { sampleIds, pageActionComponent } = this.props;
+    const { pageActionComponent } = this.props;
     // `pageActionComponent` is a render prop to add a component at the top right of the table
     // Good for a add/remove samples button. It's a function that receives the currently displayed
     // samples as an argument
--- a/src/containers/Experiment/SamplesTable.js
+++ b/src/containers/Experiment/SamplesTable.js
@@ -5,9 +5,7 @@ import 'react-table/react-table.css';
 
 import Pagination from '../../components/Pagination';
 import Dropdown from '../../components/Dropdown';
-import { RemoveFromDatasetButton } from '../Results/Result';
 import { getAllDetailedSamples } from '../../api/samples';
-import { AnonymousSubject } from 'rxjs';
 import ModalManager from '../../components/Modal/ModalManager';
 
 import './SamplesTable.scss';
@@ -24,7 +22,7 @@ export default class SamplesTable extends React.Component {
   };
 
   render() {
-    const { samples, dataSet, pageActionComponent } = this.props;
+    const { samples, pageActionComponent } = this.props;
     // `pageActionComponent` is a render prop to add a component at the top right of the table
     // Good for a add/remove samples button. It's a function that receives the currently displayed
     // samples as an argument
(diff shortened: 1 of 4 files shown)


src/containers/Experiment/SamplesTable.js
```diff
@@ -23,20 +21,16 @@ export default class SamplesTable extends React.Component {
...
-  handleAddSamplesToDataset = samples => {
-    const { accessionCode: accession_code, addSamplesToDataset } = this.props;
-    addSamplesToDataset([{ accession_code, samples }]);
-  };
-
-  handleRemoveSamplesFromDataset = sampleIds => {
-    const { accessionCode, removeSamplesFromDataset } = this.props;
-    removeSamplesFromDataset(accessionCode, sampleIds);
-  };
+  get totalSamples() {
```
ramenhog: Hmm, I don't know I've ever seen `get` in front of a class method declaration? What does this do?

src/containers/Experiment/SamplesTable.js
```diff
@@ -68,25 +57,10 @@ export default class SamplesTable extends React.Component {
                     selectedOption={this.state.pageSize}
                     onChange={this.handlePageSizeChange}
                   />
-                  of {samples.length} Samples
+                  of {this.totalSamples} Samples
```
ramenhog: In reference to my above comment, so you wouldn't need to do `this.totalSamples()`?

src/containers/Experiment/index.js
```diff
@@ -135,11 +135,37 @@ let Experiment = ({
...
                   <h2 className="experiment__title">Samples</h2>
                   <SamplesTable
-                    accessionCode={experiment.accession_code}
-                    samples={experiment.samples}
-                    addSamplesToDataset={addExperiment}
-                    removeSamplesFromDataset={removeSamplesFromExperiment}
-                    dataSet={dataSet}
+                    sampleIds={experiment.samples.map(x => x.id)}
+                    // Render prop for the button that adds the samples to the dataset
+                    pageActionComponent={samplesDisplayed =>
```
ramenhog: I like this approach! But could we could we actually make a functional component like `SampleTableActions` and pass that in here? Might make it less verbose

src/containers/Experiment/SamplesTable.js
arielsvn: It's an [object property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get). Basically allows to have a function that can be called as a field. It seemed appropriate for this case... do you think we should use a regular function here?

src/containers/Experiment/index.js
arielsvn: Yes! Good idea!

src/containers/Experiment/SamplesTable.js
ramenhog: No, that sounds really appropriate here. Didn't know about it, that's super neat ✨

arielsvn:
Thanks for reviewing @ramenhog ! I added the `SampleTableActions` component, let me know what do you think.

PR #85, opened 2018-06-14
Kurtwheeler/continuous deployment

kurtwheeler:
## Issue Number

#5 

This sets up continuous deployment for the front end. Any time a branch is merged into master with a tag starting with a `v` (as in version) we will rebuild and deploy the frontend. This way we can only deploy version commits.

## Types of changes

## Functional tests

I let the build run before adding the filters so that it would only happen on version-tagged commits to master: https://circleci.com/gh/AlexsLemonade/refinebio-frontend/11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,9 +5,11 @@ jobs:
       - image: circleci/python:3-node
     steps:
       - checkout
-      - run: pip install awscli --upgrade --user
       - run: yarn install --ignore-engines
 
       # TODO: Maybe fix this? Or at least delete this comment
       - run: CI=false yarn run build
-      - run: aws s3 sync build s3://$AWS_S3_BUCKET_PATH --delete
+      - run: pip install awscli --upgrade --user
+
+      # I'm not sure why it installs here but it does...
+      - run: ~/.local/bin/aws s3 sync build s3://$AWS_S3_BUCKET_PATH --delete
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  deploy:
+    docker:
+      - image: circleci/python:3-node
+    steps:
+      - checkout
+      - run: yarn install --ignore-engines
+
+      # TODO: Maybe fix this? Or at least delete this comment
+      - run: CI=false yarn run build
```
kurtwheeler: Without `CI=false`, all warnings become errors and it won't build. Not sure how ya'll front end folks wanna handle this.

Miserlou:
I don't know if this is a deployment bug or an application bug, but anything besides the root URL is non-accessible without going through the root entrypoint:

Ex: http://staging.refine.bio/results?q=medulloblastoma
screen shot 2018-06-14 at 10 20 46 am

I think I remember this happening in Cognoma but it looks like it's fixed there so maybe we can apply that same fix here?

Miserlou:
It'd also be great if we could get a `/robots.txt` on staging with 

```
User-agent: *
Disallow: /
```

Miserlou:
Related: https://github.com/AlexsLemonade/refinebio-frontend/pull/87

.circleci/config.yml
```diff
@@ -0,0 +1,26 @@
...
+      - run: pip install awscli --upgrade --user
+
+      # I'm not sure why it installs here but it does...
```
dongbohu: That's the default location of executable files in `pip install --user` command.

Issue #77, opened 2018-06-08
Use `accession_code` instead of `id` for samples


ramenhog:
Currently, the samples are being associated with each experiment in a dataset by their `id`. We need to change it to use their `accession_code`.

cgreene:
Is this ready for a frontend change or is something needed from the backend as well?

ramenhog:
I've been talking to Rich yesterday and this may require a backend change. I'll follow up with him again today.

PR #90, opened 2018-06-18
Ramenhog/update/77 accession codes ids swap

ramenhog:
## Issue Number
#77 

Updating the code to use `accession_code` instead of `id` to identify samples.

## Types of changes

## Functional tests
* Add/remove samples (by experiment and species) still work
* Sample tables still displays correct information on both experiment page level and downloads page level


ramenhog:
Thanks for the review! It looks like tests are failing locally currently. So I'm trying to fix those too and then the CI build.

Issue #60, opened 2018-06-06
Add Tests for the project


arielsvn:
They run, but there seems to be an error.

```
yarn run test
```

arielsvn:
Installing watchman through brew fixed the error I was getting https://github.com/facebook/react-native/issues/10088#issuecomment-261625809. With that the tests run :)

arielsvn:
![image](https://user-images.githubusercontent.com/1882507/43860964-61fccb6e-9b23-11e8-8733-bf4409f2b8c3.png)

All tests passing, we can continue writing new ones gradually. Closing this issue.

PR #93, opened 2018-06-19
fix tests locally

ramenhog:
## Issue Number
#60 

* Fix tests for `helpers.js` and `App.js` locally

## Types of changes

## Functional tests
* Tests run locally

### DIFF
--- a/src/components/Modal/ModalManager.js
+++ b/src/components/Modal/ModalManager.js
@@ -4,7 +4,7 @@ import Button from '../Button';
 
 import './Modal.scss';
 
-Modal.setAppElement('#root');
+if (process.env.NODE_ENV !== 'test') Modal.setAppElement('#root');
 
 // This class makes it easier to create a Component that displays a modal dialog.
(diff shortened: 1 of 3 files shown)


src/components/Modal/ModalManager.js
```diff
@@ -4,7 +4,7 @@ import Button from '../Button';
 
 import './Modal.scss';
 
-Modal.setAppElement('#root');
+if (process.env.NODE_ENV !== 'test') Modal.setAppElement('#root');
```
dongbohu: Does CircleCI set `process.env.NODE_ENV` to "test"?

ramenhog: When the command `yarn test` runs, it sets it to "test" apparently

dongbohu: So the test will fail with `Modal.setAppElement('#root')`?

ramenhog: Yes: https://github.com/reactjs/react-modal/issues/632

Issue #155, opened 2018-07-16
Datasets Aren't Delivered // Make The Thing Actually Work


Miserlou:
I have never actually seen a smashed and delivered dataset.

This bug is priority number one.

Miserlou:
Looks like it's trying to a) start the download before fetching the token, b) fetch the token from the wrong server, c) never use the token and d) parse some JSON that isn't JSON then fail completely.

Miserlou:
Waiting for a redeploy to test this, but hopefully it's sorted by the two associated PRs.

arielsvn:
https://github.com/AlexsLemonade/refinebio-frontend/issues/233 might be causing this.

arielsvn:
@Miserlou can we close this?? Now only processed datasets can be added for download and it seems the datasets are being delivered correctly.

PR #165, opened 2018-07-17
Actually Get And Use Token

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/155

## Types of changes
## Checklist

### DIFF
--- a/src/containers/DataSet/index.js
+++ b/src/containers/DataSet/index.js
@@ -10,6 +10,7 @@ import Button from '../../components/Button';
 import { connect } from 'react-redux';
 import { editEmail, fetchDataSet } from '../../state/dataSet/actions';
 import { startDownload } from '../../state/download/actions';
+import { Ajax, asyncFetch } from '../../common/helpers';
 import ModalManager from '../../components/Modal/ModalManager';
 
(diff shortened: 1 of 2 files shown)


src/containers/DataSet/index.js
```diff
@@ -154,14 +155,9 @@ class DatasetNoEmail extends React.Component {
...
-            const token = await (await fetch('/token/')).json();
-            await fetch(`/token/${token.id}/`, {
-              method: 'POST',
-              headers: {
-                'content-type': 'application/json'
-              },
-              body: JSON.stringify({ ...token, is_activated: true })
-            });
+            const token = await Ajax.get('/token/');
+            console.log(token);
```
kurtwheeler: This looks like a debugging line.

src/containers/DataSet/index.js
```diff
@@ -296,7 +292,7 @@ class DataSetReady extends React.Component {
 
   handleSubmit = async () => {
     if (!this.state.hasToken) {
-      const token = await (await fetch('/token/')).json();
+      const token = await (await asyncFetch('/token/')).json();
```
kurtwheeler: I think we should hold off on changing this one until we're sure what it's doing and how to test it.

PR #177, opened 2018-07-19
Fix #154

### LINKED ISSUE #154: Dataset Can Be Modified After Processing Starts
Add samples to a dataset, agree to the terms and start it processing. 

Now hit view samples. You'll see an add/remove button, which is still activated and can be used to remove samples from the displayed dataset which should be in a read only state.

Example of an in-processing dataset with zero samples:
screen shot 2018-07-16 at 10 54 43 am

Willv19:
## Issue Number

#154 

In all places in our app, we use the SamplesTable object to display lists of Samples, including on the downloading DataSet page. I introduced an isImmutable prop to SamplesTable, so in places where you would expect the dataset to be immutable it hides the 'Add/Remove' column.

## Types of changes

## Functional tests

I ran the frontend locally.

## Checklist

## Screenshots

2018-07-19-10 24 12-screenshot

### DIFF
--- a/src/containers/DataSet/index.js
+++ b/src/containers/DataSet/index.js
@@ -48,7 +48,11 @@ class DataSet extends React.Component {
               {dataSetId &&
                 (this.props.dataSet.is_processed ||
                   !!this.props.dataSet.email_address) && (
-                  <ViewDownload dataSetId={dataSetId} isEmbed={true} />
+                  <ViewDownload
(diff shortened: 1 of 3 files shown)


Willv19:
I think it's worth noting that before, the remove button did nothing but remove it from the table on that page. It did not have an affect on the current dataset.

PR #180, opened 2018-07-19
Redirect to dataset page if job was sent

### LINKED ISSUE #178: `is_processing` Datasets Should Redirect from Download to Dataset Page
You can currently view the 'Download' (modify Dataset) page for Datasets which are already processing. These should redirect to the unmodifiable dataset processing/delivery page.

To repro:

Create a new dataset
Submit the job
Press the Download Dataset Button
Press View Samples

You are presented with the option to modify and immutable dataset.

Willv19:
## Issue Number

Fixes #178 

Once we have began a smash job, it does not make sense to go back to the downloads page, since at the moment we have no concept of multiple local smash jobs. Because of that, this code redirects the user to view the status of the currently downloading dataset if the download button is pushed after a smash job was started.

## Types of changes

## Functional tests

I ran the frontend locally

## Checklist

### DIFF
--- a/src/containers/Downloads/index.js
+++ b/src/containers/Downloads/index.js
@@ -90,9 +90,10 @@ Download = connect(
       areDetailsFetched,
       samples,
       dataSet,
-      experiments
-    },
-    dataSet: { is_processing, is_processed }
+      experiments,
+      is_processing,
+      is_processed
+    }
   }) => ({
     dataSetId,
     isLoading,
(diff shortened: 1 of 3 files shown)


Miserlou:
This also needs to work on the `is_processed` field, since `is_processing` is only true _during_ the act of processing etc

Miserlou:
Better!

Issue #102, opened 2018-07-02
"Clearing" A Dataset


Miserlou:
This is probably a larger discussion, but it seems like there is no notion of "clearing" a dataset, or starting a "new" dataset. There's a usability trade off in every direction here, but the current approach of _just-keep-adding-to-the-pile_ seems like it hasn't been considered properly, and means that I have to do all of my refinebio work in an incognito tab.

dongbohu:
@dvenprasad Does it make sense to add a "Clear Dataset" button when the dataset is not empty?

dvenprasad:
![screen shot 2018-07-25 at 10 26 31 am](https://user-images.githubusercontent.com/15315514/43207600-6546c262-8ff6-11e8-8d76-41397f2372ff.png)

I think the 'Clear dataset' should be triggered on the Download page. Having a 'Remove All' button in the Samples section on the downloads page is a good way to go about this. 

Users will spend a good chunk of their time and effort to build the dataset and it shouldn't be too easy for them clear dataset accidentally, especially since we have no mechanism in place for recovery or undo. 'Remove all' button on the download page will give the users an opportunity to review what they are removing from their dataset and less likely to remove it by accident.

This fix is assuming that the dataset is automatically cleared when a smash job is kicked off at the backend.

Miserlou:
I'd like to throw in the idea of having a confirmation dialog box, as I have already deleted my dataset by mistake once.

dongbohu:
@dvenprasad: I talked to @cgreene briefly about this issue. We decided to move this button to `Download` page for now. 

I will also add a confirmation dialog box before the action.

dongbohu:
@dvenprasad It seems a little too hard for users to find this new button when it is on the right side of the `Samples` table. Does it make more sense to put it on the right side of `Share Dataset` button?

dvenprasad:
@dongbohu What is the use case are you thinking about?

dongbohu:
@dvenprasad Suppose a user adds some experiment or sample data, but for some reason he/she wants to start from scratch. I feel that the button that clears the whole dataset should be easy for the user to find. When the button is on the right side of "Samples" table, it is in the middle of the `Download` page, and it is only affiliated with samples. 

Since this button affects the whole dataset, it seems to make more sense to put it together with "Share Dataset" button. Just my two cents.

dongbohu:
Solved by PR #181.

PR #181, opened 2018-07-20
New feature: "Clear dataset" button

dongbohu:
## Issue Number
#102

* A new button that will clear the whole dataset
* Disable `has publication` filter when its counter is 0.

## Types of changes

## Checklist

## Screenshots
See the demo page on my local desktop:
http://165.123.67.202:3000/results?q=neuro

### DIFF
--- a/src/components/Checkbox/index.js
+++ b/src/components/Checkbox/index.js
@@ -1,11 +1,19 @@
 import React from 'react';
 import './Checkbox.scss';
 
-const Checkbox = ({ name, onToggle, checked, children, className }) => (
+const Checkbox = ({
+  name,
+  onToggle,
+  checked,
+  children,
+  className,
+  disabled
+}) => (
   <div className={`checkbox ${className}`}>
     <input
       type="checkbox"
       className="checkbox__input"
+      disabled={disabled}
       name={name}
       id={name}
       checked={checked}
--- a/src/containers/Downloads/DownloadBar.js
+++ b/src/containers/Downloads/DownloadBar.js
@@ -50,9 +50,7 @@ let DownloadBar = ({
       >
         {({ hideModal }) => (
           <div>
-            <h1 className="share-link-modal__title">
-              Are you sure you want to remove the dataset?
-            </h1>
+            <h1>Are you sure you want to remove the dataset?</h1>
             <div className="downloads__fieldset">
               <Button
                 buttonStyle="secondary"
--- a/src/containers/Downloads/DownloadBar.js
+++ b/src/containers/Downloads/DownloadBar.js
@@ -5,17 +5,11 @@ import ModalManager from '../../components/Modal/ModalManager';
 import InputCopy from '../../components/InputCopy';
 import './DownloadBar.scss';
 import { startDownload } from '../../state/download/actions';
-import { clearDataSet } from '../../state/download/actions';
 import { connect } from 'react-redux';
 import { getDomain } from '../../common/helpers';
 import { Link } from 'react-router-dom';
 
-let DownloadBar = ({
-  dataSetId,
-  aggregation,
-  aggregationOnChange,
-  clearDataSet
-}) => {
+let DownloadBar = ({ dataSetId, aggregation, aggregationOnChange }) => {
   return (
     <div className="downloads__bar">
       <ModalManager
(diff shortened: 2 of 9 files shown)


Willv19:
It looks like there is an issue with the navbar CSS and the buttons:
2018-07-24-15 04 36-screenshot
If I am not mistaken, this should be fixed by my latest PR, which fixed the issue of navbar `li`s not having the same width as their children. Try merging from dev and seeing if the style issue is still there.

Other than that, it looks good to me.

dongbohu:
@Willv19: The "clear" button" does l;ook much better with the latest code in `dev`. Thanks.  

@Miserlou or @Willv19: If you don't have any other comments, can you approve this PR so that I can merge it? Thanks.

dongbohu:
@Willv19 and @Miserlou: @dvenprasad and I had a short discussion about whether it will be better to move this `Clear` button to `Download` page. Let's keep this PR open for now before we decide which place will be better. Thanks.

Miserlou:
Confirming the CSS issue:

screen shot 2018-07-25 at 10 56 02 am

Looks like the functionality is there though. 

I think in it's current positioning, it is too easy to click and erase all work. I suggest that we either/or/both - moving from `[Download] [Clear]` to `[Clear] [Download]`, and adding a confirmation dialog box before we actually clear the dataset. This is probably a Deepa question.

dongbohu:
I talked to @cgreene briefly about the location of this button. I am going to move the button to `Doadload` page for now. We can decide whether this is the best choice after some user interviews later.

dongbohu:
@Miserlou, @Willv19 and @dvenprasad: I have updated the code, which has the `Remove Dataset` button on `Download` page. See the demo here:
http://165.123.67.202:3000/results?q=neuro
Just add some samples then click `Download` button on the header. Right now the button is located on the right side of `Share Dataset` button.

Miserlou:
Alright!

screen shot 2018-07-26 at 3 45 41 pm

It looks like the spacing is off on my view, the Remove button is kind of floating in the center, should the have the same spacing as the pair on the right?

I'd also suggest that we change this from [Yes] [Cancel] to  [ No, Keep This Dataset] [Yes, Clear Dataset]
screen shot 2018-07-26 at 3 45 46 pm

Miserlou:
Functionality is great though, I'm happy we have this now.

dongbohu:
@Miserlou: I have changed texts on both buttons. About the space between `Share Dataset` and `Remove Dataset` buttons, I tried a few classes, but they all put the two buttons shoulder by shoulder w/o any space, like this: 
screenshot from 2018-07-26 17-03-21

I actually prefer a larger space between the two buttons, because it makes the two buttons more distinct.

cgreene:
I'd ask @dvenprasad to weigh in.

dvenprasad:
Linking this comment from [#102](https://github.com/AlexsLemonade/refinebio-frontend/issues/102#issuecomment-408211101) here since it make sense to address it here.

In terms of what the users can see, removing all samples is equivalent to clearing the dataset.
Based on my understanding of our user's mental modeI, I think even if there's a 'remove dataset' button on top, most of the users will scroll down to look at the samples anyway. I also don't think that users will clear their dataset as often. Given that we currently have brakes (confirmation pop-up) and no mechanism for recovery, it should be with samples to minimize the likelihood of accidents.

dongbohu:
Okay, let me move the button to `Samples` section.

dongbohu:
@dvenprasad, @Miserlou and @Willv19: I have moved the `Remove All` button to `Samples` section of the `Download` page. See the demo page: http://165.123.67.202:3000/results?q=neuro
and this screenshot:
screenshot from 2018-07-27 11-52-48

As you can see, this is not the exact location that @dvenprasad specified earlier. The main reason is because the two tabs `Species View / Experiments View` and the samples table after them are an integrated component, so it is not easy to insert a separate button between them. I tried a few CSS options but they didn't work, so I gave up.

Willv19:
I like the placement, but I think the wording in the confirmation dialogue is a little bit clunky, and it might seem odd to someone who is using the website that is not clear on our concept of a 'dataset'. @dvenprasad what do you think of the wording?

dongbohu:
@Willv19 Good point. I forgot to modify the texts on both buttons after moving the location. Since they are now in `Samples` section, maybe change them to: `Yes, remove all samples` and "No, keep all samples`?

dongbohu:
I changed the wording a little bit. Please check out the demo page again.

Willv19:
LGTM

dvenprasad:
@dongbohu  That button placement is okay. I checked the padding on the button it says it's 4 x 24 but it look definitely looks bigger than that. It has a line height of 1.5 which is also. Do you think there's anything else that could be affecting this?

dongbohu:
@dvenprasad: Let me dig deeper to see why the new button looks bigger.

dongbohu:
@dvenprasad: How do you like the button now?
http://165.123.67.202:3000/results?q=neuro

dvenprasad:
Looks good! 😄

Issue #194, opened 2018-07-27
Migrate to Helmet


Willv19:
@Miserlou pointed out on Slack that the NFL has created a react component for managing the HTML `<head>` tags called Helmet: https://github.com/nfl/react-helmet. We should migrate to this to manage our `<head>` tags so that things like #104 can be implemented easier.

Miserlou:
I actually have this done in a branch, I can PR right now..

Issue #104, opened 2018-07-05
Title Is the Same on Every Page


Miserlou:
The page title only ever says 'refine.bio' - it should be updated to reflect the current page of the application.

PR #195, opened 2018-07-27
Titles

Miserlou:
## Issue Number

https://github.com/AlexsLemonade/refinebio-frontend/issues/104
https://github.com/AlexsLemonade/refinebio-frontend/issues/194

This was really just an experiment branch originally to try Helmet out, but I guess since 194 is open we might as well take a look at it. Pages which don't use the implicit render use a document.title call componentDidCreate. I think this guess all of the pages but I might have missed one.

For the record, I think it's absolutely insane that it takes this much effort just to set a page title in 2018, and that we live in a diseased and withering society which cannot be sustained for very much longer.

## Types of changes

### DIFF
--- a/src/components/Terms/Terms.js
+++ b/src/components/Terms/Terms.js
@@ -1,8 +1,12 @@
 import React from 'react';
+import Helmet from 'react';
 import './Terms.scss';
 
 const Terms = () => (
   <div className="terms">
+    <Helmet>
+      <title>refine.bio - Terms of Use</title>
+    </Helmet>
     <h1 className="terms__title">Terms of Use</h1>
     <p>
       Welcome to the Alex’s Lemonade Childhood Cancer Data Lab, which is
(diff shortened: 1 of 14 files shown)


src/components/Terms/Terms.js
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
+import Helmet from 'react';
```
Willv19: should be `from 'react-helmet'`

Willv19:
You are missing `ViewDownload`, `Experiment`, and `NoMatch` from containers. I would also move the main page title from `SearchInput` to `Main`. Also, every title can be done with `<Helmet>` tags, the tag just goes in the `render()` method of class-based components.

Willv19:
The titles in some containers should still be moved to `<Helmet>` tags.

Willv19:
LGTM!

Issue #130, opened 2018-07-06
Call to Action (CTA) when a subset of samples is already in Dataset. 


dvenprasad:
Current Behavior: 

![screen shot 2018-07-06 at 2 30 21 pm](https://user-images.githubusercontent.com/15315514/42394725-2738904e-8129-11e8-99d3-c92eb42a1b8d.png)

Expected Behavior:
It should display a button 'Add Remaining' allowing the user to add the rest of the samples to their dataset

![screen shot 2018-07-06 at 2 30 50 pm](https://user-images.githubusercontent.com/15315514/42394747-39bb9acc-8129-11e8-8e4c-9d758ccf12b6.png)

PR #199, opened 2018-07-27
Add CTA to all add buttons

### LINKED ISSUE #197: Search Page Samples Added Counter is Wrong
On my branch for the CTA issue:
2018-07-27-10 05 50-screenshot

It should say "1 sample added to dataset" at the top.

Willv19:
## Issue Number

Fixes #197 and #130 

I added a CTA to all add buttons when a subset of the samples are already in the dataset. Also, I fixed the calculation on the results page for how many samples are in the dataset from that page.

## Types of changes

## Functional tests

I ran the frontend locally

## Checklist

## Screenshots

2018-07-27-10 45 14-screenshot
2018-07-27-10 45 24-screenshot

### DIFF
--- a/src/containers/Experiment/SamplesTable.js
+++ b/src/containers/Experiment/SamplesTable.js
@@ -683,6 +683,7 @@ function AddRemoveCell({ original: sample, row: { id: rowId } }) {
             }
           ])
         }
+        buttonStyle={'secondary'}
       />
     );
   }
--- a/src/containers/Experiment/SamplesTable.js
+++ b/src/containers/Experiment/SamplesTable.js
@@ -12,7 +12,10 @@ import FileIcon from './file.svg';
 import ProcessIcon from './process.svg';
 import { PAGE_SIZES } from '../../constants/table';
 import SampleFieldMetadata from './SampleFieldMetadata';
-import { RemoveFromDatasetButton } from '../../containers/Results/Result';
+import {
+  RemoveFromDatasetButton,
+  AddToDatasetButton
+} from '../../containers/Results/Result';
 import {
   addExperiment,
   removeExperiment,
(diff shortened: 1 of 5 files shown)


dvenprasad:
The 'Add Remaining' button needs to be a ghost button.

Mockup:
screen shot 2018-07-27 at 10 59 22 am

Miserlou:
This is good, I like these buttons a lot more!

I was able to get it into a state where there are two of the same button button on the search page, I'm not sure if this is the intended behavior or not:

Miserlou:
I was able to generate a crash on the download page:

screen shot 2018-07-27 at 11 12 16 am

And there are add buttons on my dataset preview page:
screen shot 2018-07-27 at 11 13 32 am

Clicking them causes the crash

Willv19:
2018-07-27-11 21 51-screenshot

dvenprasad:
- The add buttons in the table are supposed to be ghost buttons
- It's overkill having how many samples are added on a per page basis. Also, the CTA there is ambiguous. Am I adding all samples or only samples on the page?  I would revert back to original CTA 'Add Page to Dataset' without the information block(i.e X samples already in dataset).  Ideally, the users should be able use the sort on the add/remove column to get look at the added samples.

Willv19:
Is this how it should look?
2018-07-31-12 07 51-screenshot

dvenprasad:
Yep, looks good!

Miserlou:
I am still able to repro the crash by removing a sample, reloading the downloading page, then pressing 'Add' on the view samples view.

Willv19:
I agree that it's a problem, but I was also able to reproduce the same crash on the staging server, so it can't be something added by this PR. We should probably file a separate ticket for it.

Miserlou:
Ah, okay. Will file a ticket and approve this.

Issue #141, opened 2018-07-10
Cannot navigate forward on browser


dvenprasad:
From the search results page, navigate to an experiment page, then hit the back button on the browser. This takes the user back to the search page but cannot use the forward button on the browser to go back to the experiment page.

arielsvn:
For the record: This bug was caused because we always update the browser url after every search. This is done by calling `history.push`. When navigating back from an experiment page, this update is triggered, which adds a new entry in the browser history removing the entry for the experiments page, and disabling the forward arrow.

This is the default behavior of `history.push`, adding a new entry in the browser history even though there won't be any change in the url. There's some discussion around it, such as https://github.com/ReactTraining/history/issues/470. 

I used some of the suggestions and modified the behavior of `history` in https://github.com/AlexsLemonade/refinebio-frontend/pull/230/commits/604b269914b1e57f445f15ff08db4ea8dd576cf3

PR #230, opened 2018-08-08
Fix #185

### LINKED ISSUE #185: Filters change based on filtered results isn't correct
When the user adds a filter to the search results, ALL filters are reloaded and only the filters applicable on the filtered results is visible. This limits how users can filter. 

For example, say the user wants to filter results based on human AND mouse. They cannot do that with the way filtering currently behaves. When they choose mouse, all filters are reloaded based on the filtered results, taking away the option to also filter by Human.

Before applying a filter:
screen shot 2018-07-23 at 4 13 32 pm

After:
screen shot 2018-07-23 at 4 13 40 pm

Expected Behavior:
In the above example, when user applies the filter on organism, the filter options should change for Technology and Publication but all options should remain for Organism.

### LINKED ISSUE #244: Refactor: Use the url as the source of truth for search filters
In the search page, we're parsing the filters from the url and copying them into the Redux store. Then when a filter is applied or a new search term is entered, the changed is saved into the Redux store and then pushed into the url.

We can simplify the code if we use the url as the single source of truth for the filters and the search term, instead of having to maintain the store and the url synced.

### LINKED ISSUE #246: When all filters in a category is applied, it redirects to the 'No results found' page
Current Behavior:
When I applied both microarray and rna-seq as a filter, it took me to 'No matching results' page. 

Expected Behavior:

- In this case, it should still display all results.

- If there are no results, due to the number of filters applied, then it should display: (Currently, if I were to hit the back button from this, it will take me back to the landing page and not to the state before.) 

screen shot 2018-08-15 at 3 07 37 pm

Willv19:
## Issue Number

close #185
close #244 
close #246 

The way we handled multiple filters in the same category before was comma-separating the two options, which was okay because we never had the backend hooked up to do this. Now, the backend requires both options in their own key-value pair with the same key, so this changes the frontend code to account for that.

## Types of changes

## Functional tests

I ran the frontend locally against an instance of the API running the new changes.

## Checklist

## Screenshots

2018-08-08-14 16 34-screenshot
2018-08-08-14 16 37-screenshot
2018-08-08-14 16 40-screenshot
2018-08-08-14 16 43-screenshot


Willv19:
Relies on https://github.com/AlexsLemonade/refinebio/pull/503 getting merged and deployed first.

src/common/helpers.js
```diff
@@ -84,7 +84,12 @@ export function formatSentenceCase(str) {
 
 // Helper methods to ease working with ajax functions
 export const Ajax = {
-  get: (url, params = false) => {
+  get: (url, params = false, stringParams = []) => {
+    if (stringParams.length > 0) {
+      return asyncFetch(
+        `${url}?${getQueryString(params)}&${stringParams.join('&')}`
```
arielsvn: what about passing the values in `stringParams` as key/values of `params` ??

src/state/search/actions.js
```diff
@@ -27,27 +27,25 @@ export function fetchResults(searchTerm = '', pageNum = 1, filters) {
        */
       filtersObj = Object.keys(appliedFilters).reduce((result, filterType) => {
```
arielsvn: I like this use of `reduce` 💪

src/common/helpers.js
Willv19: Because we are allowing more than one filter of the same category to be applied at the same time, `params` might get instantiated something like this: `'organism__name': 'HOMO_SAPIENS', 'organism__name': 'GALLUS_GALLUS'`, but keys in an object in javascript must be unique so the last filter to be applied overwrites the other filters. The stringParams is a workaround to allow multiple filters with the same key to be added to the URL

arielsvn: I see, what if we modify the `getQueryString` function to process array values in that way?? So that we can do:

```
Ajax.get('/url', {'organism_name': ['HOMO_SAPIENS', 'GALLUS_GALLUS']}
```

It feels kind of strange that `get` has two parameters to receive the url arguments.

src/common/helpers.js
```diff
@@ -6,7 +6,19 @@
 export function getQueryString(queryObj) {
   return Object.keys(queryObj)
     .filter(key => queryObj[key] !== undefined)
-    .map(key => `${key}=${encodeURI(queryObj[key])}`)
+    .reduce((accum, key) => {
+      // For some query parameters, the key points to an array of values.
+      // In those instances we want a separate param for each value,
+      // otherwise we just want the one param for the value.
+      if (typeof key === 'string') {
```
arielsvn: Nice Refactor! but shouldn't we check here for the type of `queryObj[key]` ? And also make sure it works with other types besides `string`.

dongbohu: Ariel: The staging server's API is not back yet, so I can't really test the web UI, but I think you are right. Another thing that confuses me is the second and third screenshots. In the second screenshot, when the first organism is selectd, why the URL doesn't have this search parameter? Same thing with the third screenshot when both organisms are selected.

Another issue that I noticed was that `getQueryString` was imported in `src/state/search/actions.js` but was never used. So that import line should be removed.

dongbohu: @arielsvn  can you modify the code to address issues that we found out in this PR?

arielsvn: @dongbohu thanks for reviewing. I'll address these comments later today.

arielsvn:
@Miserlou @dongbohu this is ready to be reviewed now.

> Another thing that confuses me is the second and third screenshots. In the second screenshot, when the first organism is selectd, why the URL doesn't have this search parameter? Same thing with the third screenshot when both organisms are selected.

@dongbohu It seems that problem was introduced in https://github.com/AlexsLemonade/refinebio-frontend/pull/207. I added a fix here, and also for https://github.com/AlexsLemonade/refinebio-frontend/issues/141

dongbohu:
@arielsvn Here is the demo server that is built based on this branch:
http://165.123.67.202:3000/results?q=GSE
I have a few questions (some of them may not be related to this PR):
1. The numbers inside `()` for each filter seems to be the number of samples, but I remember it used to be the number of matched experiments. Wouldn't it make more sense to use the number of experiments for each filter?
 
2. If you check any filters, the result is updated, but why the URL doesn't change?

3. If you type in another search term, the result is updated, but the URL is still unchanged. Why?

dongbohu:
@arielsvn Thank you for the update. I rebuilt the demo server and tried the web UI. It behaves better but still not perfectly. On my demo server, say we have this URL:
http://165.123.67.202:3000/results?q=GSE
If I check the first organism filter `Danio rerio`, the result is updated, and the URL becomes:
http://165.123.67.202:3000/results?q=GSE&organisms__name=DANIO_RERIO
This is good. But when I also check another organism filter `Homo Sapiens`, 
For a very short moment the URL became:
http://165.123.67.202:3000/results?q=GSE&organisms__name=DANIO_RERIO&organisms__name=HOMO_SAPIENS
but it eventually becomes: 
http://165.123.67.202:3000/results?q=GSE&organisms__name=HOMO_SAPIENS
So somewhere in the code the location path is not synced correctly with the actual query parameters.

arielsvn:
I see, good catch @dongbohu! I think I know what's causing this, it's probably related with the method that parses the url now that we have two parameters with the same name.

I'll push a fix later and let you know. Thanks!

dongbohu:
@arielsvn, I put more thoughts on this feature and here is what I would expect:
(suppose we have `http://hostname/results?q=GSE`):

1. When user selects an organism filter say `org1`:
    1) Only show results whose organism is `org1`
    2) URL becomes: `http://hostname/results?q=GSE&organisms__name=org1`

2. When user selects another organism `org2` (but still keeps `org1` selected):
    1) Only show results whose organisms are either `org1` or `org2`
    2) URL becomes: `http://hostname/results?q=GSE&organisms__name=org1,org2`

3. When user also selects one of `Technology` filters, say `tech1`:
    1) Only show resutls whose organisms are `org1` OR `org2`, AND whose technology is `tech1`
    2) URL becomes: `http://hostname/results?q=GSE&organisms__name=org1,org2&technology=tech1`

4. When user types in another search term, say `foo`:
    1) Show `foo` search results whose organisms are `org1` OR `org2`, AND whose technology is `tech1`
    2) URL becomes: `http://hostname/results?q=foo&organisms__name=org1,org2&technology=tech1`

In `2.ii`, the original implementation seems to (want to) change the URL to:
`http://hostname/results?q=GSE&organisms__name=org1&organisms__name=org2`
I think this is a little misleading, because it actually means `organisms__name=org1` OR `organisms__name=org2`, not `organisms__name=org1` AND `organisms__name=org2`. And (pun intended) we are using `&` when our filters are in different categories, such as `organisms__name=org1&technology=tech1`.

Please let me know whether this expectation makes sense. 

(Please also forgive me if I have been sounding too picky on this PR.)

arielsvn:
> In 2.ii, the original implementation seems to (want to) change the URL to:
http://hostname/results?q=GSE&organisms__name=org1&organisms__name=org2
I think this is a little misleading, because it actually means organisms__name=org1 OR organisms__name=org2, not organisms__name=org1 AND organisms__name=org2. And (pun intended) we are using & when our filters are in different categories, such as organisms__name=org1&technology=tech1.

It makes sense, in fact we're using commas when separating other parameters such as sample's accession codes in the samples endpoint. The filters feature was implemented on the backend by https://github.com/AlexsLemonade/refinebio/pull/503. I think we might have to modify it to be able to process filters how you propose.

If you check this [test case](https://github.com/AlexsLemonade/refinebio/blob/8f0ea7077441d980eff709d32e106345b12d98b2/api/data_refinery_api/tests.py#L366) they use `technology=MICROARRAY&technology=FAKE-TECH` instead of `technology=MICROARRAY,FAKE-TECH` as you suggest.

Initially I didn't like the idea of using `organisms__name=org1&organisms__name=org2` with the same parameter name multiple times in the url. But I [read online](https://stackoverflow.com/a/2602127/763705) this is accepted by the HTTP standard. So both that, and `organisms__name[]=org1&organisms__name[]=org2` are accepted ways of sending `[org1, org2]` to the server.

That said, if you prefer we can modify the filters to send `organisms__name[]=org1,org2` instead.



> 4. When user types in another search term, say `foo`:
    1) Show `foo` search results whose organisms are `org1` OR `org2`, AND whose technology is `tech1`
    2) URL becomes: `http://hostname/results?q=foo&organisms__name=org1,org2&technology=tech1`

I'm not sure I agree with this point, in my opinion if the user changes the search query we should restart the filters, since they depend on the search term. The filters `organisms__name=org1,org2` may not make sense for the term `foo`.

If you search something on amazon for example, they do this, they clear the filters after each search.

> (Please also forgive me if I have been sounding too picky on this PR.)

Don't worry, it's important that we get this right and I enjoy talking about this :)

dongbohu:
> I'm not sure I agree with this point, in my opinion if the user changes the search query we should restart the filters, since they depend on the search term. The filters organisms__name=org1,org2 may not make sense for the term foo.

You are definitely right. I forgot that the filters would be shown dynamically depending on the search results.  So please reset all filters when user types in a new search term.

I didn't know that `organisms__name=org1&organisms__name=org2` in the URL is accepted by HTML standard. Let's keep this format then. I will get used to it.

Thanks.

arielsvn:
@dongbohu this should be ready for review again. I apologize in advance, I had to do a couple changes to get the urls working properly (this will close #244 too). I also realized that we were doing two identical requests to the server to fetch the results, and fixed it here.

Let me know what do you think. Thanks!

arielsvn:
This will also close https://github.com/AlexsLemonade/refinebio-frontend/issues/246

src/containers/Results/index.js
```diff
@@ -1,38 +1,275 @@
...
-
-const ResultsContainer = connect(mapStateToProps, {
+  }), {
   ...resultsActions,
   ...downloadActions
 })(Results);
 
-export default ResultsContainer;
+export default Results;
+
```
dongbohu: I think one or two blank lines would be good enough. Any reason why there are 5 here?

arielsvn: I agree, it seems I hadn't configured correctly vscode to apply prettify to the code. I pushed changes now to fix this. Thanks!

arielsvn:
Also fixed conflicts that appeared after merging https://github.com/AlexsLemonade/refinebio-frontend/pull/240

Issue #192, opened 2018-07-26
Download Page Is Extremely Slow for Large Datasets


Miserlou:
This may be a backend issue, or it may be a frontend-not-using-the-API-properly issue.

[...]

PR #236, opened 2018-08-10
Make the downloads page faster

Willv19:
## Issue Number

#192 

Fix the frontend so that it expects the new traffic returned from the backend after these changes were made: https://github.com/AlexsLemonade/refinebio/pull/517.

## Types of changes

## Functional tests

I ran the API and frontend locally.

## Checklist

### DIFF
--- a/src/api/dataSet.js
+++ b/src/api/dataSet.js
@@ -5,14 +5,12 @@ export async function getDataSet(dataSetId) {
 }
 
 export async function getDataSetDetails(dataSetId) {
-  return await Ajax.get(`/dataset/${dataSetId}/`, {details: true});
+  return await Ajax.get(`/dataset/${dataSetId}/`, { details: true });
 }
 
 export async function updateDataSet(dataSetId, dataSet) {
   return await Ajax.put(`/dataset/${dataSetId}/`, { data: dataSet });
 }
-
-
 
 // Takes in arrays of samples and experiments as formatted by the serializer
 // and turns them into objects where the keys are experiment accession codes
@@ -27,7 +25,7 @@ export function formatExperiments(experiments) {
 export function formatSamples(data, samples) {
   let result = {};
 
-  for (let accession_code in data) {
+  for (let accession_code of Object.keys(data)) {
     result[accession_code] = samples.filter(sample =>
       data[accession_code].includes(sample.accession_code)
     );
@@ -35,4 +33,3 @@ export function formatSamples(data, samples) {
 
   return result;
 }
-
--- a/src/api/dataSet.js
+++ b/src/api/dataSet.js
@@ -4,6 +4,10 @@ export async function getDataSet(dataSetId) {
   return await Ajax.get(`/dataset/${dataSetId}/`);
 }
 
+export async function getDataSetDetails(dataSetId) {
+  return await Ajax.get(`/dataset/${dataSetId}/?details=true`);
+}
+
 export async function getSamplesAndExperiments(dataSet) {
   const experiments = {},
     samples = {};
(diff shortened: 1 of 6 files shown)


Willv19:
Relies on https://github.com/AlexsLemonade/refinebio/pull/517 landing

src/common/helpers.js
```diff
@@ -114,3 +114,25 @@ export const getMetadataFields = experiment =>
   SampleFieldMetadata.filter(field =>
     experiment.sample_metadata.includes(field.id)
   ).map(field => field.Header);
+
+// Takes in arrays of samples and experiments as formatted by the serializer
+// and turns them both into objects where the keys are experiment accession
+// codes
+export const formatSamplesAndExperiments = (data, samples, experiments) => {
```
arielsvn: What about splitting this into two methods? `formatSamples` and `formatExperiments`

src/state/download/actions.js
```diff
@@ -246,26 +246,68 @@ export const fetchDataSetSucceeded = (
...
 };
 
+export const fetchDataSetDetailsSucceeded = (
+  dataSet,
+  is_processing,
+  is_processed,
+  aggregate_by,
+  scale_by,
+  samples,
+  experiments
```
arielsvn: It might be better to use [an object](https://github.com/AlexsLemonade/refinebio-frontend/pull/51#discussion_r192516899) as a parameter

dongbohu: I agree, since the input arguments are exactly same as the value of `data` in the returned object.

src/api/dataSet.js
```diff
@@ -4,36 +4,35 @@ export async function getDataSet(dataSetId) {
...
+  return experiments.reduce((accum, experiment) => {
+    accum[experiment.accession_code] = experiment;
+    return accum;
+  }, {});
+}
+
+export function formatSamples(data, samples) {
+  let result = {};
+
+  for (let accession_code in data) {
```
dongbohu: I don't feel very comfortable about the `for/in` loop here, because it will iterate inherited properties too:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in
Although it is not wrong to use `for/in` here, I would prefer to iterate over `Object.keys(data)`, which will include direct properties only. What do you think, @arielsvn?

arielsvn: You're right @dongbohu, thanks for reviewing! I'll update this later and let you know

arielsvn:
This should be ready for review. Thanks @dongbohu !

src/state/download/actions.js
arielsvn: Ups, I missed this. Will upload the change in a sec, sorry!

arielsvn:
@dongbohu I fixed conflicts and addressed your last comment. Let me know if you spot anything else. Thanks!

src/state/download/actions.js
dongbohu: When you said `use an object as a parameter`, I thought you meant this:
```
export const fetchDataSetDetailsSucceeded = dataObj => { type: 'DOWNLOAD_FETCH_DETAILS_SUCCESS', ...dataObj }
```
Am I too crazy?

arielsvn: hahah nah I don't think so... This is an interesting approach that I hadn't considered, although in my opinion, in this case listing the variables that we're passing helps with the readability of the code. We write these variables in the reducer as well, just to emphasize what we're passing https://github.com/AlexsLemonade/refinebio-frontend/blob/5d4e6c5c21a64d77324a2337b377f2ba644f1d30/src/state/download/reducer.js#L70-L79

In any case, both would be equivalent, and yours is definitely shorter! I can change it if you want.

PR #238, opened 2018-08-10
Clear the current dataset after starting a download

### LINKED ISSUE #231: Create a new dataset after you start downloading the current one
When a dataset starts being processed it becomes immutable, and the user gets redirected to the dataset specific page every time the download button is clicked. So after the download starts the app becomes unusable, since users are not able to add/remove samples or create a new dataset.

Talked to @dvenprasad and the current dataset should be removed when a download is started, that way users will be able to start creating a new download.

arielsvn:
## Issue Number

Close #231 

Removes the current dataset after a download starts. 

## Types of changes

## Functional tests

I started a new download.

## Checklist

## Screenshots

download reset

### DIFF
--- a/src/state/download/actions.test.js
+++ b/src/state/download/actions.test.js
@@ -1,7 +1,6 @@
-
-import {fetchDataSet, startDownload} from './actions';
-import configureMockStore from 'redux-mock-store'
-import thunk from 'redux-thunk'
+import { fetchDataSet, startDownload } from './actions';
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
 import { REPORT_ERROR } from '../reportError';
 
 const middlewares = [thunk];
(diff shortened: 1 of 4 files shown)


Miserlou:
Doesn't this change mean that the user will never see the Processing or Download states? Shouldn't it be cleared when the file is _downloaded_, rather than submitted for processing?

arielsvn:
Not necessarily, the user can still see the processing and download states in the dataset specific page, which will be opened automatically after a download is started.

http://localhost:3000/dataset/88697da2-8cfd-4e8c-86ac-a48567be9d80

The only difference is that the current dataset will be removed, so the user can click the download button which redirects to `/downloads`.

arielsvn:
@Miserlou I added a screenshot of how it's supposed to work, let me know what do you think. This is ready to be reviewed now.

src/state/download/actions.test.js
```diff
@@ -39,7 +39,25 @@ describe('fetchDataSet', () => {
 
     await store.dispatch(fetchDataSet());
 
-    expect(store.getActions().map((x)=> x.type)).toEqual(["DOWNLOAD_DATASET_FETCH", "DOWNLOAD_CLEAR", "DOWNLOAD_CLEAR_SUCCESS", REPORT_ERROR]);
+    expect(store.getActions().map((x)=> x.type)).toEqual(["DOWNLOAD_DATASET_FETCH", "DOWNLOAD_CLEAR", REPORT_ERROR]);
```
dongbohu: Trivial question: It seems that most of the front-end code adds a space before and after `=>`. Can you also add a space before `=>` here?

arielsvn: Sure! I'm don't know why this wasn't picked by prettify

PR #240, opened 2018-08-15
Prevents unprocessed samples from being added to the dataset

### LINKED ISSUE #233: Unprocessable Samples Are Displayed And Can Be Added to Datasets
I've only processed 3 of these samples so far, but all 102 are showing up:

screen shot 2018-08-09 at 3 15 13 pm

In the API, these have `is_processed: False`

screen shot 2018-08-09 at 3 16 37 pm

This will lead to people being able to create Datasets that will never successfully process as they expect.

arielsvn:
## Issue Number

close #233 

This won't allow users to add unprocessed samples to the dataset. When implementing this I added several refactors to simplify the actions that add/remove samples from the current dataset, and also some tests.

The buttons in the search results to add samples won't work, since the application is expecting to have information about the sample `is_processed` state. I filed a ticket  for this https://github.com/AlexsLemonade/refinebio/issues/529

## Types of changes

## Functional tests

Added removed several samples from the experiments page.
Check the downloads page that only processed samples are added.

## Checklist

## Screenshots

add to dataset


cgreene:
This pull request **fixes 14 alerts** when merging 2b5ac464ea6bfd9da48b792e8ab3445af5f9e35f into 20dd0c931bfb1311d13fbcb3cf5c3c42ac6ac503 - [view on LGTM.com](https://lgtm.com/projects/g/AlexsLemonade/refinebio-frontend/rev/pr-cd808e85fe0949891bd31b0d76608c13847546d2)

**fixed alerts:**

* 12 for Unused variable
* 1 for Unused loop iteration variable
* 1 for Useless assignment to local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

src/containers/Downloads/downloadFilesData.js
```diff
@@ -63,15 +63,10 @@ function estimateGeneExpressionSize(dataSet) {
   let totalSize = 0;
 
   // calculate the size of each one of the experiments
-  for (let experimentId of Object.keys(dataSet)) {
+  for (let experimentId in dataSet) {
```
dongbohu: I can see why the original version used `for/of` iteration because it is safer. If you are sure that dataSet is never an object that inherits from other object, then I guess `for/in` loop is okay.

src/containers/Experiment/DataSetSampleActions.js
```diff
@@ -0,0 +1,174 @@
...
+        buttonStyle="secondary"
+        onClick={handleAdd}
+      />
+      <p className="dataset-add-button__info-text">
+        <i className="ion-information-circled dataset-add-button__info-icon" />{' '}
+        {totalSamplesInDataset} Samples are already in Dataset
+      </p>
+    </div>
+  );
+}
```
dongbohu: New line character needed here.

src/containers/Experiment/ProcessingInformationCell.js
```diff
@@ -0,0 +1,394 @@
...
+import React from 'react';
+import Button from '../../components/Button';
+import ModalManager from '../../components/Modal/ModalManager';
+
+import FileIcon from './file.svg';
+import ProcessIcon from './process.svg';
+
+
+// Pipeline names ref: https://github.com/AlexsLemonade/refinebio-frontend/issues/22#issuecomment-394010812
+const PIPELINES = {
```
dongbohu: With the introduction of `Processor` introduced in each `ComputationalResult` record, this variable may have to be updated. See:
https://github.com/AlexsLemonade/refinebio/blob/2e48ea9b16d2b5a3b0ffa074d28052800a746571/common/data_refinery_common/models/models.py#L410
We probably should figure out a better way to synchronize these constants than hardcoded strings. Can you file a new issue to keep track of it? Thanks.

src/containers/Experiment/ProcessingInformationCell.js
```diff
@@ -0,0 +1,394 @@
+
```
dongbohu: This file seems to be moved from `src/containers/Experiment/SamplesTable.js`, right? If it is, I don't have to review it line by line.

src/state/download/reducer.js
```diff
@@ -52,14 +52,7 @@ export default (state = initialState, action) => {
...
     }
-    case 'DOWNLOAD_REMOVE_EXPERIMENT_SUCCESS': {
-      const { dataSet } = action.data;
-      return {
-        ...state,
-        dataSet
-      };
-    }
-    case 'DOWNLOAD_REMOVE_SPECIES_SUCCESS': {
+    case 'DOWNLOAD_UPDATE_DATASET': {
```
dongbohu: Just curious: why some actions has a "xxx_SUCCESS" affiliated but some don't? Thanks.

src/containers/Experiment/ProcessingInformationCell.js
arielsvn: Sure, I added the issue here https://github.com/AlexsLemonade/refinebio-frontend/issues/249 for further discussion.

src/containers/Experiment/ProcessingInformationCell.js
arielsvn: That's right, I didn't do any modifications there

src/state/download/reducer.js
arielsvn: I think the main reason is when we need to track the state of a request in the redux store. For example, if we needed a resource named `FOO`, we could have the actions `FETCH_FOO` and `FETCH_FOO_SUCCESS`. Here we were associating two actions for each request that we needed to do to the server.

In this specific case, I guess this action could also have been called `DOWNLOAD_UPDATE_DATASET_SUCCESS`, since it's called mostly after a request to the server that modifies the dataset. However I saw it as something more general, we can use it every time that we want to update the dataset in the store.

This is not the only way to do it, in some cases it's simpler to track the state of the requests on the views. For that we have the component [Loader](https://github.com/AlexsLemonade/refinebio-frontend/blob/dev/src/components/Loader/index.js). I wanted to start doing refactors to gradually simplify the store and the action creators, and probably remove some of these actions.

Do you think this makes sense??

dongbohu: Thanks @arielsvn. It makes sense. :+1:

arielsvn:
I addressed the comments @dongbohu, let me know if you see anything else. Thanks!

PR #260, opened 2018-08-22
Add Species filters on downloads page

### LINKED ISSUE #251: On download page> experiments view, users can't filter experiments based on species
Current:
screen shot 2018-08-16 at 11 31 00 am

Expected:
screen shot 2018-08-16 at 11 31 19 am

arielsvn:
## Issue Number

close #251 

Add filters for species in the downloads page.

## Types of changes

## Functional tests

Checked downloads page, and toggled the filters

## Checklist

### DIFF
--- a/src/components/Radio/Radio.scss
+++ b/src/components/Radio/Radio.scss
@@ -1,6 +1,6 @@
 @import '../../common/styles/variables';
 
-.checkbox {
+.radio {
   display: flex;
   align-items: center;
   margin-top: 8px;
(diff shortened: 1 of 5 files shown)


arielsvn:
After comments from @dvenprasad I pushed changes to list all possible species.

image

Miserlou:
The change is functional, but I think it really confuses the functionality of the whole section which seems like a total mess now.

If anything, this view is now powerful enough that the "Species View" can be dropped entirely, since Experiment:All Species is practically the same thing. (If we really wanted, we could display an "View All Samples" button on that view, but I don't think that'd be necessary). Having the toggle switcher seems super clunky now.

arielsvn:
Makes sense, I'm not sure about removing the species view, since that groups the samples differently. It seems like it can be improved. Tagging @dvenprasad

Thanks for reviewing!

PR #273, opened 2018-08-28
Arielsvn/support/235 calls localstorage refactor

### LINKED ISSUE #235: Refactor: Remove the calls to `localStorage` from the action creators
We're storing the id of the current dataset that the user is creating in the `localStorage`, and modifying this value in several places in the action creators. It seems un-necessary, to have a parallel `localStorage` besides the redux store. 

Another approach could save the `dataSetId` in the redux store, and keep this value in sync with the `localStorage` with [redux-persist](https://github.com/rt2zz/redux-persist) or a similar solution. That way the action creators would only have to interact with the redux store.

arielsvn:
## Issue Number

close #235 

Removes calls to localStorage from the action creators, these we're moved to a single place in `configureStore.js`.

## Types of changes

## Functional tests

Created/removed datasets for download.

## Checklist

### DIFF
--- a/src/configureStore.js
+++ b/src/configureStore.js
@@ -74,15 +74,15 @@ function routerMiddleware(history) {
 
 /**
  * Loads the state from the localStorage, here are the keys that we're interested in persisting.
- * In the future if this get's more complicated, we can consider using https://github.com/rt2zz/redux-persist
+ * In the future if this gets more complicated, we can consider using https://github.com/rt2zz/redux-persist
  * For now, just loading these keys here and persisting them in `store.subscribe` should be good
  * enough.
  *
  * Inspired from https://egghead.io/lessons/javascript-redux-persisting-the-state-to-the-local-storage
  *
- * In general it might not be a good idea to have to store know about `dataSetId` which is something
+ * In general it might not be a good idea to have the store know about `dataSetId` which is something
  * specific of a reducer. In the future we could add a new reducer with the data that the application needs
- *
+ * saved
  */
 function loadInitialState() {
   return {
--- a/src/configureStore.js
+++ b/src/configureStore.js
@@ -42,7 +42,7 @@ const store = createStore(
 store.subscribe(
   throttle(() => {
     const state = store.getState();
-    if (state.download.dataSetId) {
+    if (state.download && state.download.dataSetId) {
       localStorage.setItem('dataSetId', state.download.dataSetId);
     } else {
       localStorage.removeItem('dataSetId');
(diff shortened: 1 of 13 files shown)


src/configureStore.js
```diff
@@ -5,22 +5,11 @@ import thunk from 'redux-thunk';
 import history from './history';
 import { CALL_HISTORY_METHOD } from './state/routerActions';
 import { REPORT_ERROR } from './state/reportError';
+import throttle from 'lodash/throttle';
```
dongbohu: Interesting. I never used `lodash` before. It is even possible to install throttle only: https://www.npmjs.com/package/lodash.throttle
I am not sure whether it applies to our case though.

src/configureStore.js
```diff
@@ -70,7 +65,29 @@ function routerMiddleware(history) {
...
+    const {
+      payload: { method, args }
+    } = action;
     history[method](...args);
   };
 }
+
+/**
+ * Loads the state from the localStorage, here are the keys that we're interested in persisting.
+ * In the future if this get's more complicated, we can consider using https://github.com/rt2zz/redux-persist
```
dongbohu: `get's` to `gets`?

src/configureStore.js
```diff
@@ -70,7 +65,29 @@ function routerMiddleware(history) {
...
+ * For now, just loading these keys here and persisting them in `store.subscribe` should be good
+ * enough.
+ *
+ * Inspired from https://egghead.io/lessons/javascript-redux-persisting-the-state-to-the-local-storage
+ *
+ * In general it might not be a good idea to have to store know about `dataSetId` which is something
```
dongbohu: Is `have to store know ...` supposed to be `have the store know ...`?

src/containers/Downloads/DownloadDetails.js
```diff
@@ -229,7 +230,7 @@ class ExperimentsView extends React.Component {
                       {metadataFields && metadataFields.length ? (
                         metadataFields.join(', ')
                       ) : (
-                        <i class="result__not-provided">
+                        <i className="result__not-provided">
```
dongbohu: Good catch! I am surprised it was not reported by babel-loader.

src/configureStore.js
arielsvn: Yes it's possible, although it's the same, importing `lodash/throttle` only adds the `throttle` function. And we have the `lodash` package in case we need another function from there. I think I also used `isEqual` in a different place.

src/containers/Downloads/DownloadDetails.js
arielsvn: yes, I think I saw it with LGTM or Lint

src/configureStore.js
dongbohu: Good to know. Thanks!

arielsvn:
Just fixed the typos @dongbohu, thanks for pointing those out.

Issue #99, opened 2018-06-29
Get organism genome build and version information dynamically for sample processing modals


ramenhog:
Relies on this issue: https://github.com/AlexsLemonade/refinebio/issues/163

jaclyn-taroni:
I edited the title to reflect the fact that we will have different genome build information for every individual organism

Willv19:
@dvenprasad @jaclyn-taroni Where on the frontend do we want this information?

jaclyn-taroni:
The processing information modals, see: https://github.com/AlexsLemonade/refinebio-frontend/issues/22 and https://github.com/AlexsLemonade/refinebio-frontend/pull/50. Let us know if that doesn't clarify things!

arielsvn:
@dongbohu is there an api for this?? I noticed that https://github.com/AlexsLemonade/refinebio/pull/313#issuecomment-398153073 wasn't merged since the approach to solve this changed.

Also found https://github.com/AlexsLemonade/refinebio/pull/346 which seems to update the models.

dongbohu:
@arielsvn Yes, the API is has been written but it is not completely working yet. I am working on a PR that fixes the backend for this API. When that PR is done, the API should work.

arielsvn:
Great to hear @dongbohu! Please ping me when that PR is done to finish this issue.

arielsvn:
@dongbohu I was checking in staging after the last deploy... I think the samples processor information is not being returned correctly. Check [here](https://api.staging.refine.bio/samples/?accession_codes=GSM414029) for example.

It's returning a number instead of the processor information

![image](https://user-images.githubusercontent.com/1882507/44158190-24e54d00-a082-11e8-9e89-6888d1f71839.png)

I'm missing something?

dongbohu:
I think `1` is the processor ID. You can get more details of processors by:
https://api.staging.refine.bio/processors/

I will modify the backend to support the API like `https://api.staging.refine.bio/processors/1`.

arielsvn:
> I will modify the backend to support the API like https://api.staging.refine.bio/processors/1.

In the future there will be a big number of processors?? I see we only have 5 on staging right now. We can also do a single request to `/processors` and save that information in the front end, to avoid having to make individual requests for each processor. What do you think?

dongbohu:
Right now we have 11 types of processors (some of them are not created in the database because they are not being used). So it is feasible to get all processors in one request and cache them on the front end. 

[...]

Issue #225, opened 2018-08-07
Submitter-processed protocol information appears to be hardcoded


jaclyn-taroni:
All submitter-processed protocol information appears to be the same, regardless of the experiment in question.

`GSM182283` from `GSE7529` 

![screen shot 2018-08-07 at 2 35 31 pm](https://user-images.githubusercontent.com/19534205/43795457-62c90302-9a4f-11e8-8e9a-31672a576b43.png)

`GSM399040` from `E-GEOD-15892`

![screen shot 2018-08-07 at 2 37 47 pm](https://user-images.githubusercontent.com/19534205/43795481-7c30e4f4-9a4f-11e8-99c5-1ad2079af551.png)

I believe this may be hardcoded?

https://github.com/AlexsLemonade/refinebio-frontend/blob/c15940459fdd39072e5c05089129f4d06990550d/src/containers/Experiment/SamplesTable.js#L349

We'll need to supply the protocol information for the sample that the user has clicked on. The information on [this page](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM399040) is what I would expect to see for `GSM399040`.

arielsvn:
Tagging @Miserlou. We'll probably need a new field on the `/samples` endpoint with the *Submitter Supplied Protocol*.

dongbohu:
@Ariel, just want to clarify, for sample `GSM399040`, on the modal's "Submitter Supplied Protocol", you want to show all the information listed on https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM399040 or only some fields?

jaclyn-taroni:
The fields that I think are important there are `Extraction protocol`, `Label protocol`, `Hybridization protocol`, `Scan protocol` and `Data processing`.

dongbohu:
@jaclyn-taroni Do you happen to know whether the backend already retrieves and saves the information on this page:
 https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM399040 
in the database? I checked the code but didn't find it.

cgreene:
Seems likely to be a @Miserlou question to me. I think he wrote that bit.

jaclyn-taroni:
Also I can say, from looking at the metadata I got as part of the jackiecrunch, it appears to have protocol information in the sample annotations?

jaclyn-taroni:
Question for @dvenprasad ☝️

[...]

PR #280, opened 2018-08-30
Generate Processing information dialogs dinamically

### LINKED ISSUE #249: Create Processing Information dialogs dynamically
From https://github.com/AlexsLemonade/refinebio-frontend/pull/240/files/2b5ac464ea6bfd9da48b792e8ab3445af5f9e35f#r210742381

> With the introduction of Processor introduced in each ComputationalResult record, this variable may have to be updated. See:
https://github.com/AlexsLemonade/refinebio/blob/2e48ea9b16d2b5a3b0ffa074d28052800a746571/common/data_refinery_common/models/models.py#L410
We probably should figure out a better way to synchronize these constants than hardcoded strings. Can you file a new issue to keep track of it? Thanks.

All of the information regarding the Pipelines and the Processing information dialogs is stored statically in the front-end. With the introduction of `Processor` we might also be able to build these dialogs dynamically. I have to look more into this to see how it could be done.

This api endpoint https://github.com/AlexsLemonade/refinebio-frontend/issues/99#issuecomment-413269352 will probably be helpful.

### LINKED ISSUE #265: Copy changes for processing information modals
While writing documentation, I came across some tweaks I wanted to make to the copy for the processing information modals.  I'm going off of what we have on AlexsLemonade/refinebio#170 for the RNA-seq pipeline because I can't check this on staging at the moment (https://github.com/AlexsLemonade/refinebio-frontend/issues/249#issuecomment-413974258).

#### Affymetrix SCAN

> the SCAN.UPC bioconductor package documentation (DOI: 10.18129/B9.bioc.SCAN.UPC).

That DOI link currently goes to the publication URL, it should be `https://doi.org/10.18129/B9.bioc.SCAN.UPC`

#### tximport

`gene-level count values` -> `gene-level count-scale values`

#### Submitter Processed

There are two different gene identifier conversion pipelines depending on the microarray manufacturer, so we'll need to update the copy to reflect that:

> We have created custom gene mapping files for Affymetrix platforms (see: https://github.com/AlexsLemonade/identifier-refinery) that support conversion from probe IDs, gene symbols, Entrez IDs, RefSeq and Unigene identifiers to Ensembl gene IDs.
> We support conversion from Illumina BeadArray probe IDs to Ensembl gene IDs using Bioconductor Illumina BeadArray expression packages.

arielsvn:
## Issue Number

#99 
close #249 
close #265 

This adds changes to generate the processing information dialogs based on the `ComputationalResults` returned by the samples. From there we take the pipelines that were applied and also their versions.

In this PR we're hiding *MultiQC*.

Also the samples were returning computational results that were repeated. In this PR I only took one of each, and also ordered them by the creation date.

https://github.com/AlexsLemonade/refinebio-frontend/blob/cbebd3a46825c67e967239fc30e834e7a02a737f/src/containers/Experiment/ProcessingInformationCell.js#L20-L31

Missing from this PR is fetching the Genome information dynamically #99, and also the submitter processed protocol information #225.

## Types of changes

## Functional tests

Tested the dialog on several samples, both on the experiments page and on the downloads page.

## Checklist

## Screenshots

image

image

image


arielsvn:
Pushed changes to also fetch the Genome build dynamically, so this should also close #99

src/containers/Experiment/ProcessingInformationCell.js
```diff
@@ -4,319 +4,259 @@ import ModalManager from '../../components/Modal/ModalManager';
...
+      </div>
+    );
+  }
+
+  static PROTOCOLS = {
+    'Affymetrix SCAN': AffymetrixScanProtocol,
+    'Illumina SCAN': IlluminaScanProtocol,
+    'Transcriptome Index': null,
+    'Salmon Quant': SalmonProtocol,
+    MultiQC: null,
```
dongbohu: Let's enclose `MultiQC` and `Tximport` with ' ' too to keep the format consistent.

src/containers/Experiment/ProcessingInformationCell.js
```diff
@@ -4,319 +4,259 @@ import ModalManager from '../../components/Modal/ModalManager';
...
+    Tximport: TxtimportProtocol,
+
+    'Submitter-processed': SubmitterSuppliedProtocol
+  };
```
dongbohu: Do we also need a `SalmonTools` item?

src/containers/Experiment/ProcessingInformationCell.js
```diff
@@ -4,319 +4,259 @@ import ModalManager from '../../components/Modal/ModalManager';
...
+      <h3>SCAN</h3>
       <p>
-        The gene identifiers were detected and converted to Ensembl gene IDs
-        using our{' '}
+        SCAN (Single Channel Array Normalization) is a normalization method for
+        single channel (Affymetrix) microarrays that allows us to process
+        individual samples. SCAN models and corrects for the effect of technical
+        bias, such as GC content, using a mixture-modeling approach. For more
+        information about this approach, see the primary publication (Piccolo,
+        et al. Genomics. 2012.
```
dongbohu: Not sure whether we need to add more formats to this paragraph (such as use italic font for `Genomics. 2012`). Any thoughts, @jaclyn-taroni?

src/containers/Experiment/ProcessingInformationCell.js
arielsvn: I did it initially, but prettify changed it. By default object keys with a single word don't have `''`.

We could try to change the configuration of prettify, but it might not be worth it, since we have been using it everywhere till now.

src/containers/Experiment/ProcessingInformationCell.js
arielsvn: Which one is `SalmonTools` ?

dongbohu: `Salmontools` is one of the processors used on backend. I am not sure whether the front end has specific protocols for it. If not, forget about my previous comment. What do you think, @jaclyn-taroni?

src/containers/Experiment/ProcessingInformationCell.js
dongbohu: Never mind then. Sometimes Prettify is too smart :(

src/containers/Experiment/ProcessingInformationCell.js
jaclyn-taroni: Not yet, similar to `MultiQC`

src/containers/Experiment/ProcessingInformationCell.js
jaclyn-taroni: +1 _Genomics._ 2012.

We'll probably want to be consistent in italicizing journal names in all protocols, then.

src/containers/Experiment/ProcessingInformationCell.js
arielsvn: I agree, I don't particularly like how they organize the html either, but I have gotten more used to it

PR #283, opened 2018-09-04
Fix download workflow

### LINKED ISSUE #279: Use new email field to determine if a dataset has an email associated
For security reasons https://github.com/AlexsLemonade/refinebio/commit/94059ba9270a72478bcf56a5a1eb55c7d9a6d823 removed the email field from the dataset endpoint.

Chatted with @Miserlou, and we'll add a new flag to the dataset endpoint to know whether a dataset has an email assigned or not.

arielsvn:
## Issue Number

close #279 

Avoids using the field `email_address` that was returned by the dataset endpoint and instead relies on the state of the url to pass this value in the download workflow.

Also included several refactors to the Download/DataSet components, like unifying the shared download page, that had it's own component, and the dataset page. Now they are a single component that can be accessed on `/dataset/{id}`.

## Types of changes

## Functional tests

Try the complete download workflow, from adding a sample to the dataset to downloading it.

Regenerate an expired download, this one for example http://localhost:3000/dataset/c9994ec5-a35e-42d3-b964-624f079acd42

## Checklist


arielsvn:
Added a new commit here to show the error message when starting a download fails:

image

close #286

PR #285, opened 2018-09-05
Add author to landing page sample search terms

### LINKED ISSUE #243: Search by author as an example search term on the landing page?
Since refine.bio can now support search by author, could we have an author show up as an example along with 'Notch, Medulloblastoma, and GSE16476'. 

Tagging @jaclyn-taroni for input.

arielsvn:
## Issue Number

Close #243 

Add author search term to landing page.

## Screenshots

image


Miserlou:
Can Notch be changed to NOTCH or NOTCH1? Every single time all I think about is [this guy](https://en.wikipedia.org/wiki/Markus_Persson)..

PR #303, opened 2018-09-13
Arielsvn/support/responsive error pages

arielsvn:
## Issue Number

N/A

Optimize error pages for small devices

## Screenshots

image

image

### DIFF
--- a/src/containers/AppError/index.js
+++ b/src/containers/AppError/index.js
@@ -35,7 +35,7 @@ let AppError = ({ goBack }) => (
   </div>
 );
 AppError = connect(
-  () => ({}),
+  null,
   {
     goBack
   }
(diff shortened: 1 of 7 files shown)


src/containers/AppError/index.js
```diff
@@ -26,12 +26,19 @@ let AppError = ({ goBack }) => (
...
+        className="img-responsive"
+      />
     </div>
   </div>
 );
-AppError = connect(() => ({}), {
-  goBack
-})(AppError);
+AppError = connect(
+  () => ({}),
```
dongbohu: This arrow function `() => ({})` is not doing anything, can we simply use `null`?

arielsvn: good idea! I wasn't sure of this, but double checked that it's possible in the [connect docs](https://github.com/reduxjs/react-redux/blob/master/docs/api.md#arguments)

arielsvn:
@dongbohu I replaced the empty arrow function for `null` everywhere in the app. There were 6 other usages. Nice catch!

PR #308, opened 2018-09-17
Filters still displayed when there are no results

### LINKED ISSUE #307: Keep filters if search doesn't render any results
Right now we don't support changing the current filters if the search didn't return anything.

Design here https://projects.invisionapp.com/d/main#/console/13557531/302436522/preview

arielsvn:
## Issue Number

close #307 

The filters will stay visible if the user applies a combination that doesn't have any results.

## Types of changes

## Functional tests

Search and apply several filters.

## Checklist

## Screenshots

Also added empty state for the search:

image

### DIFF
--- a/src/containers/Results/index.js
+++ b/src/containers/Results/index.js
@@ -19,6 +19,8 @@ import { Link } from 'react-router-dom';
 import DataSetSampleActions from '../Experiment/DataSetSampleActions';
 import isEqual from 'lodash/isEqual';
 import Loader from '../../components/Loader';
+import Button from '../../components/Button';
+import { clearFilters } from '../../state/search/actions';
 
 class Results extends Component {
   state = {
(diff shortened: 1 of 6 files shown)


src/containers/Results/index.js
```diff
@@ -132,54 +86,95 @@ class Results extends Component {
...
-              ) : null}
-
-              <DataSetSampleActions
-                data={samplesAsDataSet}
-                enableAddRemaining={false}
-                meta={{
-                  buttonStyle: 'secondary',
-                  addText: 'Add Page to Dataset'
-                }}
+        <Loader
```
dongbohu: Not really for this PR: I found the rendering logic here kind of confusing. Maybe we should add some comments to explain it?

src/containers/Results/index.js
```diff
@@ -242,46 +237,82 @@ NumberOfResults = connect(
...
+          src={GhostSampleImage}
+          alt="Start searching"
+          className="results__no-results-image img-responsive"
+        />
+      </div>
+    </div>
+  </div>
+);
+NoSearchResultsTooManyFilters = connect(
+  () => ({}),
```
dongbohu: This line can be replaced by `null`.

arielsvn:
Addressed the comments @dongbohu, let me know if we should try to explain more the rendering logic of the search results

PR #311, opened 2018-09-19
Fix dashboard

### LINKED ISSUE #140: Dashboard Is Broken
It appears to work for the first load, but every subsequent action breaks it for all time views.

screen shot 2018-07-10 at 2 19 24 pm

arielsvn:
## Issue Number

close #140 

Ensures that the dashboard only triggers an update, after 5secs have passed since the previous update. Before we had the problem described in https://github.com/AlexsLemonade/refinebio-frontend/issues/140#issuecomment-409573880

If https://github.com/AlexsLemonade/refinebio/pull/638 get's merged, the `/stats` endpoint will contain all the information required to render the dashboard. This PR adds refactors to use that information instead of calculating it in the front-end.

## Types of changes

## Functional tests

open the dashboard, select several ranges.

## Checklist

## Screenshots

image


arielsvn:
Updated the title, this should be ready for review.

src/components/TimeRangeSelect/index.js
```diff
@@ -9,29 +9,23 @@ type Props = {
...
-        </label>
-        <Field
-          className="time-range-select__dropdown"
-          name="timeRange"
-          options={props.options}
-          component={Select}
-          onChange={handleChange}
-        />
-      </div>
+let TimeRangeSelect = ({ selectedTimeRange, options }: Props) => (
```
dongbohu: Just curious, is this line equivalent to:
```let TimeRangeSelect = ({ selectedTimeRange, options }) => (```
(The destructure syntax in ES6+ confuses me from time to time).

arielsvn: Yes those are equivalent. It took me a while to get used to it too.

dongbohu: Thanks. I am going to approve this PR now. Please feel free to close it after resolving the conflicts.

arielsvn: Thanks! I'll merge it after the one in the backend.

dongbohu: By the way, since they are equivalent, do you mind removing `: Props` part from that line? I am afraid the next time when I see this line, I would still ask myself what it means.

dongbohu: I forgot that `Props` is the type annotation. You can leave it as is.

PR #335, opened 2018-09-27
add quantile normalization info to processing info dialogs

### LINKED ISSUE #330: Add quantile normalization information to processing modal
As stated in the [quantile normalization section](https://github.com/AlexsLemonade/refinebio-docs/blob/master/docs/main_text.md#quantile-normalization) of our documentation:

> Note that only gene expression matrices that we are able to successfully quantile normalize will be available for download.

As a result, all gene expression matrices delivered to a user will be quantile normalized. Although this is information is in our documentation, there is no indication of this on our processing modal at the moment. I think a short blurb is sufficient. Something like:

_All data available for download will be quantile normalized. For more information regarding how quantile normalization is performed and its limitations, see [our documentation](https://github.com/AlexsLemonade/refinebio-docs/blob/master/docs/main_text.md#quantile-normalization)._

@dvenprasad and @cgreene thoughts?

arielsvn:
## Issue Number

close #330 

@dvenprasad 

## Screenshots

image

image

### DIFF
--- a/src/containers/Experiment/ProcessingInformation/ProcessingInformationModalContent.js
+++ b/src/containers/Experiment/ProcessingInformation/ProcessingInformationModalContent.js
@@ -96,6 +96,20 @@ export default class ProcessingInformationModalContent extends React.Component {
           this._getProtocolDescription(name)
         )}
 
+        <div>
+          All samples available for download will be <b>quantile normalized</b>.
+          For more information regarding how quantile normalization is performed
+          and its limitations, see{' '}
+          <a
+            href="https://github.com/AlexsLemonade/refinebio-docs/blob/master/docs/main_text.md#quantile-normalization"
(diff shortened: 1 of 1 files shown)


Miserlou:
This is hardcoded - should this be implemented in a way that supports non-QN data? Will we _ever_ deliver non-QN data, @jaclyn-taroni ?

jaclyn-taroni:
That is definitely within the realm of possibility, yes.

Miserlou:
Okay. The problem then is that decision would occur at smash time, but this is shown when they're added to the dataset. They haven't been QN'd at that point, so I don't know if this absolute language is correct here.

jaclyn-taroni:
It is correct right now. We didn't have any indication outside the documentation that this quantile normalization was occurring.

Issue #346, opened 2018-09-28
Individual Pages Link to ID, Not Accession


Miserlou:
![screen shot 2018-09-28 at 2 59 53 pm](https://user-images.githubusercontent.com/139987/46228154-3bdcb780-c32f-11e8-9de8-7a761b1c7c3b.png)

arielsvn:
I just notices that the API also uses the ID to get details on an experiment:

https://api.refine.bio/experiments/16/

Is this alright?

arielsvn:
We could also use

https://api.refine.bio/experiments/?accession_code=SRP087617

But that doesn't return all the experiment's information.

PR #147, opened 2018-07-13
Ramenhog/bugfix/129

ramenhog:
## Issue Number
Closes #129 

## Purpose/Implementation Notes
* fix anchor to sample tables section from results

## Types of changes

What types of changes does your code introduce?
_Put an `x` in the boxes that apply_

* [x] Bugfix (non-breaking change which fixes an issue)

## Functional tests
* Now when you click on the "View samples" button on results, it takes the user to the experiment page and anchors the user at the samples table

## Screenshots
issue-129

PR #353, opened 2018-10-01
Improve experiments page loading speed

arielsvn:
## Issue Number

N/A

In the process of implementing https://github.com/AlexsLemonade/refinebio-frontend/issues/346 I added some improvements to the experiments url page.

Before we were adding `?ref=search` to the url to signal that the previous page was the search page, and show the back to search button. Here this info is sent in the url state.

Also included there information about the current search result, so that we're able to render the first part of the experiments page instantly. Since this info is pretty much the same that we display in the search results.

This introduces a strange behavior in the scrolling of the experiments page, since it now has content on the initial render and the browser tries to restore the scroll position. I have to investigate if it might be worth bringing back the `Anchor` component removed in https://github.com/AlexsLemonade/refinebio-frontend/pull/147

## Types of changes

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist

## Screenshots

This is limiting the network speed:

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
     "react-modal": "^3.4.5",
     "react-redux": "^5.0.7",
     "react-router-dom": "^4.2.2",
-    "react-router-hash-link": "^1.2.0",
     "react-table": "^6.8.6",
     "recharts": "^1.0.0-beta.10",
     "redux": "^3.7.2",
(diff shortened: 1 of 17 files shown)


arielsvn:
Fixed the scroll issue and a few cosmetic details, this should be ready for review.

Issue #378, opened 2018-10-09
Bug when multiple experiments share the same samples


arielsvn:
Discovered this bug locally, when this happens samples are not added correctly to the dataset and/or are counted incorrectly in the top right corner.

PR #379, opened 2018-10-09
Big refactor to consider experiments sharing the same samples

### LINKED ISSUE #370: Allow for datasets to be edited in multiple tabs
Right now they get into an invalid state.

arielsvn:
## Issue Number

close https://github.com/AlexsLemonade/refinebio-frontend/issues/378
close #370 

Discovered #378 in my local computer, with some experiments that shared a lot of the same samples. The operations to add/remove elements to the dataset were only taking into account the sample ids, so adding/removing samples from the search page was taking the dataset into an invalid state.

This PR adds refactors for all methods related to modifying the dataset, and also adds a few tests.

In addition it fixes #370 by updating the current dataset before any PUT requests to `/dataset`.

## Types of changes

## Functional tests

Add/remove samples on search/experiments/download pages.

## Checklist

## Screenshots

Note when the second experiment is added the number of samples in the dataset doesn't change.


src/containers/Experiment/DataSetStats.js
```diff
@@ -1,52 +1,120 @@
...
+   * eg:
+   * sameItems([1,2], [2,1]) === true
+   * @param {Array} s1
+   * @param {Array} s2
+   */
+  static _sameItems(s1, s2) {
+    return difference(s1, s2).length === 0 && difference(s2, s1).length === 0;
+  }
+
+  static equal(d1, d2) {
```
dongbohu: So this function tells whether two input objects have same keys and values? I wonder whether an implementation using `Map` in ES6 would be easier.

src/containers/Experiment/DataSetStats.js
```diff
@@ -1,52 +1,120 @@
...
+    const d1Keys = Object.keys(d1);
+    const d2Keys = Object.keys(d2);
+    if (!DataSetOperations._sameItems(d1Keys, d2Keys)) {
+      // d1 and d2 don't have the same accession codes
+      return false;
+    }
+
+    for (let accession of d1Keys) {
```
dongbohu: If this function may be used in other scenarios,  you probably should use a more generic variable name (such as `element` or `entry`) than `accession`.

src/containers/Experiment/SamplesTable.js
```diff
@@ -370,7 +361,17 @@ function ThComponent({ toggleSort, className, children, ...rest }) {
 }
 
 function AddRemoveCell({ original: sample, row: { id: rowId } }) {
-  const { experimentAccessionCode } = sample;
+  // retrieve all experiment accession codes referencing this sample
+  const { experimentAccessionCodes } = sample;
+  // Create a dataset slice, where we include all experiments that are referencing this sample
+  // that way when it get's added/removed it will impact all those experiments
```
dongbohu: Typo: `get's` should be `gets`. You may also want to add a comma or semicolon at the end of the previous line.

src/state/download/actions.js
```diff
@@ -23,37 +23,31 @@ export const downloadUpdateDataSet = dataSet => {
...
-export const removeExperiment = accessionCodes => async (
-  dispatch,
-  getState
-) => {
-  const { dataSet, dataSetId } = getState().download;
-  const newDataSet = new DataSetManager(dataSet).removeExperiment(
-    accessionCodes
-  );
+const dataSetUpdateOperation = modifier => async (dispatch, getState) => {
+  // Update the current dataset with whatever in on the server, otherwise it might get out of sync
```
dongbohu: Is `in` supposed to be `is`?

src/containers/Downloads/DownloadDetails.js
```diff
@@ -137,46 +141,55 @@ const SpeciesSamples = ({
...
-              {formatSentenceCase(speciesName)} Samples
-            </h2>
-            <div className="downloads__sample-stats">
-              <p className="downloads__sample-stat">
-                {species[speciesName].length}{' '}
-                {species[speciesName].length > 1 ? 'Samples' : 'Sample'}
-              </p>
+      {Object.keys(samplesBySpecies).map(speciesName => {
+        // Create a slice with the samples for each specie
+        // TODO: in the future we could consider refactoring `groupSamplesBySpecies`, seems unnecessary having to modify the shape of that object here
```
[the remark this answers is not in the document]
arielsvn: I think prettier didn't modify it, because starts with `TODO:`. I'll update it.

src/containers/Downloads/DownloadDetails.js
```diff
@@ -137,46 +141,55 @@ const SpeciesSamples = ({
   removeSamples,
   isImmutable = false
 }) => {
```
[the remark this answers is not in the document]
arielsvn: Yes, although prettier likes to put JSX inside `( )` for some reason. I'll change it to the lambda function directly.

src/containers/Experiment/DataSetStats.js
```diff
@@ -1,52 +1,120 @@
...
+        d2[experimentAccession]
+      );
+    }
+    return result;
   }
 
-  anyProcessedSamples() {
-    return this.samples.some(x => x.is_processed);
+  /**
+   * Returns true if both given arguments have the same values, without taking into cosideration their order
```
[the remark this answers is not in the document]
arielsvn: thanks for pointing out the typo! Please let me know if you see any others :)

src/containers/Experiment/DataSetStats.js
arielsvn: What would you suggest? It should return true if both inputs have the same keys, and values without taking into account their order. Eg if:

```
d1 = { e1: ['s1', 's2']}
d2 = { e1: ['s2', 's1']}
```

`equal(d1, d2)` should be `true`

src/containers/Experiment/DataSetStats.js
arielsvn: Maybe we can change the function name, or add comments to make this more clear. 

This function it's only meant to be used to compare `dataSets`. Which are objects, where the keys are experiment accession codes and the values are lists of sample accession codes (It's the same value we send on `data` for the `/dataset` endpoint). That's why I named this variable `accession`, to try to make it more explicit to what it was referring to.

Do you think this makes sense?

src/containers/Experiment/DataSetStats.js
dongbohu: I took a look at the methods in ES6 Map:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
it is disappointing. I thought it offered some methods to compare two Map instances directly but I was wrong. I don't think `Map` or `Set` will make the implementation easier. Please feel free to leave the code as is.

src/containers/Experiment/DataSetStats.js
dongbohu: Since the function is only used to compare datasets, using `accession` is acceptable. I just want to clarify.

arielsvn:
I addressed the comments @dongbohu, thanks for reviewing! This PR had a lot of changes

PR #387, opened 2018-10-11
Show error when the dataset processing fails

### LINKED ISSUE #377: Show dataset error as returned by the API
https://github.com/AlexsLemonade/refinebio/issues/691 added two new fields with information of any errors that might occur when a dataset is being processed. We have to update the front-end to account for that.

arielsvn:
## Issue Number

close #377 

This shows any error message returned from the backend when there's an error processing a dataset, and allows the users to re-try the process.

I wasn't able to get a dataset to fail locally, so I tested this setting the values manually in the front-end.

```
  dataSet.success = false;
  dataSet.failure_reason =
    'Failed Kolmogorov Smirnov test! Stat: 0.6254109910756225, PVal: 0.0';
```

This needs https://github.com/AlexsLemonade/refinebio/commit/0a4f00f441de1276c95bfb44882ea8462b422d22 to be deployed in order to work, but it won't fail if it isn't.

## Types of changes

## Functional tests

Check a dataset that failed when it was being processed.

## Checklist

## Screenshots

image

### DIFF
--- a/src/containers/Downloads/DataSet/index.js
+++ b/src/containers/Downloads/DataSet/index.js
@@ -146,11 +146,15 @@ function DataSetPageHeader({ dataSetId, email_address, hasError, dataSet }) {
     is_available,
     expires_on,
     s3_bucket,
-    s3_key
+    s3_key,
+    success
   } = dataSet;
 
+  // success can sometimes be `null`
+  const processingError = success === false;
+
   const isExpired = moment(expires_on).isBefore(Date.now());
-  return hasError ? (
+  return hasError || processingError ? (
     <DataSetErrorDownloading dataSetId={dataSetId} dataSet={dataSet} />
   ) : is_processed ? (
     is_available && !isExpired ? (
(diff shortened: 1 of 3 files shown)


src/containers/Downloads/DataSet/index.js
```diff
@@ -198,6 +196,34 @@ let DataSetErrorDownloading = ({
...
+                  or{' '}
+                  <a
+                    href="https://github.com/AlexsLemonade/refinebio/issues/new"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    className="link"
+                  >
+                    file a ticket on Github
+                  </a>{' '}
+                  with the below error message:
```
dongbohu: `following` is probably better than `below`. (I am not a native speaker of English either, you can double check with an expert like @jaclyn-taroni.)

src/containers/Downloads/DataSet/index.js
```diff
@@ -198,6 +196,34 @@ let DataSetErrorDownloading = ({
...
+                <span>
+                  {' '}
+                  or{' '}
+                  <a
+                    href="https://github.com/AlexsLemonade/refinebio/issues/new"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    className="link"
+                  >
+                    file a ticket on Github
```
dongbohu: Maybe this line shoud be changed to
```
report the issue to us
```
("file a ticket on GitHub" sound so techy and even scary to a user who never uses github).

arielsvn:
I edited the copy @dongbohu, let me know what do you think.

image

Issue #64, opened 2018-06-06
Landing page summary statistics


arielsvn:
Needs https://github.com/AlexsLemonade/refinebio/issues/768

---

https://github.com/AlexsLemonade/refinebio-frontend/issues/59 Won't include the Summary statistics for the landing page. We should discuss what graphs to include in that section, and where the data is coming from.

![image](https://user-images.githubusercontent.com/1882507/41043924-01ffe278-6973-11e8-9129-42e893dc5b97.png)

arielsvn:
Talked with Deepa about these graphs.  We'll try to do something similar to https://stripe.com/us/sigma with the tabs of the graphs.

![2018-10-18 13 39 14](https://user-images.githubusercontent.com/1882507/47173131-40b3dc00-d2db-11e8-8d13-6146d643174b.gif)

Here're some details for each graph specifically:

**Samples per species**

We'll have it as a bar graph, with the top 10-20 organisms, whatever looks good depending on the user resolution. Initially I thought we'll need to add an extra endpoint for this, but maybe we can grab this information from the filters in the search endpoint.

![image](https://user-images.githubusercontent.com/1882507/47172729-49f07900-d2da-11e8-9ea5-08fdcddbff15.png)

Each bar graph should link to the search page.

**Samples over time**

This will be a line graph, similar to the one we have in the dashboard. I was thinking we can use the same endpoint. https://github.com/AlexsLemonade/refinebio/issues/707 will help with the performance.

**Volume of data over time**

Is this something we can calculate/estimate? @Miserlou @kurtwheeler

arielsvn:
https://github.com/AlexsLemonade/refinebio-frontend/pull/410 Added the graphs "Samples per Species" and "Samples over Time", but the latter one is very slow since it uses the `/stats` endpoint from the API.

https://github.com/AlexsLemonade/refinebio/issues/766 and https://github.com/AlexsLemonade/refinebio/issues/768 on the backend should provide the support neccesary to complete this issue.

arielsvn:
Closing this, we'll be able to add the volume of data downloaded over time graph once https://github.com/AlexsLemonade/refinebio/issues/768 is addressed.

PR #410, opened 2018-10-31
Landing page summary stats

arielsvn:
## Issue Number

#64 

This adds two graphs with stats to the landing page: Samples per Species and Samples over Time.

Clicking on one of the bars from the Samples per Species graph takes the user to a search for all experiments matching that organism.

The Samples over time graph is using the `/stats` endpoint from the api, this is the same we use for the dashboard, and it's very slow. I'll file an issue in the backend to see how we can improve it.

## Types of changes

## Checklist

## Screenshots

image
image


Miserlou:
The stats endpoint was really not intended for this kind of usage, and we might not always want it to be public.. it might be better if these numbers were hard-coded. It doesn't seem to be too bad performance wise right now but.. =/

arielsvn:
Added a script to fetch the data we need from `/stats` on every deploy (@kurtwheeler 's idea), now the graphs work fine.

PR #420, opened 2018-11-07
Verify deploys

### LINKED ISSUE #416: Only allow specific users to push tags that trigger deploys
Similar to what is done in refine.bio

https://github.com/AlexsLemonade/refinebio/blob/dev/.circleci/verify_tag.sh

arielsvn:
## Issue Number

close #416 

Only allow authorized people to trigger deploys.

## Types of changes

## Functional tests

We should be able to test this on staging.


.circleci/verify_tag.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This script verifies that the tag triggering this deploy was signed
```
dongbohu: `deploy` should be `deployment`.

.circleci/verify_tag.sh
```diff
@@ -0,0 +1,12 @@
...
+# by a trusted member of the CCDL.
+
+for key in $(ls -1 keys/); do
```
dongbohu: This line can be simplified into:
```for key in keys/*; do```

File: .circleci/verify_tag.sh (hunk 1 above)
Comment (arielsvn) [reply .circleci/verify_tag.sh]:
```suggestion
# This script verifies that the tag triggering this deployment was signed
```

(trying github's suggestion feature)

.circleci/verify_tag.sh
arielsvn:
```suggestion
for key in keys/*; do
```

arielsvn:
I addressed the comments @dongbohu! and got to try a new github's feature. They now allow to make code suggestions in the comments

dongbohu:
Interesting. Never heard of this feature!

kurtwheeler:
:raised_hands: :raised_hands: :raised_hands:

PR #433, opened 2018-11-21
Api optimizations

arielsvn:
## Issue Number

https://github.com/AlexsLemonade/refinebio/pull/861

## Purpose/Implementation Notes

https://github.com/AlexsLemonade/refinebio/pull/861 introduces breaking changes to the API. This PR contains updates to handle these changes.

## Types of changes

* [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Functional tests

Test search and experiments page.

## Checklist

* [x] Lint and unit tests pass locally with my changes
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] I have added necessary documentation (if appropriate)
* [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots

image

PR #451, opened 2018-12-03
Reduce the amount of information that is sent with `pushState`

### LINKED ISSUE #436: Clicking View Samples Does Nothing
screen shot 2018-11-26 at 7 58 39 pm

Firefox https://www.refine.bio/results?q=

arielsvn:
## Issue Number

Close #436 

When navigating to the experiments page, we are sending the result object attached to the state of the URL. Firefox [has a limit of 640kb](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState()_method). 

Some experiments were bigger than 640kb, and that was making the navigation fail in FF.

Once https://github.com/AlexsLemonade/refinebio-frontend/pull/433 get's merged these objects will be smaller.

/cc @Miserlou 

## Types of changes

## Functional tests

Search and navigate to experiments page.

## Checklist

### DIFF
--- a/src/containers/Results/Result/index.js
+++ b/src/containers/Results/Result/index.js
@@ -37,7 +37,18 @@ const Result = ({ result, query }) => {
             className="link result__title"
             to={routes.experiments(result.accession_code, {
               ref: 'search',
-              result
+              result: {
+                accession_code: result.accession_code,
+                title: result.title,
+                description: result.description,
+                pubmed_id: result.pubmed_id,
+                publication_title: result.publication_title,
+                submitter_institution: result.submitter_institution,
+                publication_authors: result.publication_authors,
+                source_url: result.source_url,
+                source_database: result.source_database,
+                samples: []
+              }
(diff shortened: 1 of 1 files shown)


src/containers/Results/Result/index.js
```diff
@@ -37,7 +37,18 @@ const Result = ({ result, query }) => {
             className="link result__title"
             to={routes.experiments(result.accession_code, {
               ref: 'search',
-              result
+              result: {
```
dongbohu: So here some attributes are removed from `result`, and a new `samples` attribute is added?

arielsvn: Yes, basically to avoid adding too much information in the url state. 

The samples attribute will be removed with https://github.com/AlexsLemonade/refinebio-frontend/pull/433, once that's merged the API won't send any detailed sample information. I added a default value here to avoid changing the experiments page here since that's updated on #433

Issue #481, opened 2018-12-21
Samples table broken on downloads page for large datasets


arielsvn:
When there're more than 2k samples, we need to calculate the samples on the backend instead of sending all the accession codes in the URL.

Similar to https://github.com/AlexsLemonade/refinebio-frontend/issues/121#issuecomment-404906582

PR #485, opened 2019-01-08
Refactor download page after API changes

### LINKED ISSUE #472: Download Page Continuously Makes Gigantic Requests for Large Datasets
I'm not sure why this is. FF.

screen shot 2018-12-18 at 11 20 25 am

Loading this page for the first time is very slow, which seems very unnecessary as when you View Samples, they don't load instantly (if at all).

arielsvn:
## Issue Number

close #472 
close https://github.com/AlexsLemonade/refinebio-frontend/issues/481

https://github.com/AlexsLemonade/refinebio/pull/962 added updates for the download page endpoints. This PR accounts for those changes in the frontend.

## Types of changes

## Functional tests

Test downloads page workflow.

## Checklist

## Screenshots

Additions to the dataset endpoint when `/?details=true`

image

### DIFF
--- a/src/api/dataSet.js
+++ b/src/api/dataSet.js
@@ -14,9 +14,13 @@ export async function getDataSetDetails(dataSetId) {
 }
 
 export async function updateDataSet(dataSetId, dataSet, details = false) {
-  // Q: is it wrong to include a query parameter in a PUT request?
-  // the `details` parameter is only used to "modify" the reponse from the API, it doesn't
-  // cause any side effect or different behavior
+  // It's not technically wrong to add a query parameter to the URL of a PUT request. ref https://github.com/AlexsLemonade/refinebio-frontend/pull/485#discussion_r246158557
+  // In this case, after a dataset is modified we'll need to fetch detailed information for it,
+  // which will include the samples grouped by organism for example.
+  // For a fully restfull api we could do a sepparate request for those, but that seems unnecessary
+  // since we can have that calculated on this request.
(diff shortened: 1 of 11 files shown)


arielsvn:
Also included here a fix for #481, we won't send the sample accession codes anymore, instead we'll use the following parameters:

- for species view: `/samples/?offset=0&limit=10&dataset_id=...&organism__name=DANIO_RERIO`

- for experiments: `/samples/?offset=0&limit=10&dataset_id=...&experiment_accession_code=GSE37642`

src/api/dataSet.js
```diff
@@ -9,13 +9,25 @@ export async function getDataSetDetails(dataSetId) {
...
     experiments: formatExperiments(response.experiments)
   };
 }
 
-export async function updateDataSet(dataSetId, dataSet) {
-  return await Ajax.put(`/dataset/${dataSetId}/`, { data: dataSet });
+export async function updateDataSet(dataSetId, dataSet, details = false) {
+  // Q: is it wrong to include a query parameter in a PUT request?
+  // the `details` parameter is only used to "modify" the reponse from the API, it doesn't
+  // cause any side effect or different behavior
```
Miserlou: This is a great question, the answer is no, it's not technically wrong, although it looks a bit confusing because normally the query would be what's set by the PUT. You should change this comment to explain this incongruous behavior. It isn't RESTful though and we can change the API behavior if necessary to clarify this.

PR #486, opened 2019-01-09
Support stacked modal dialogs

### LINKED ISSUE #444: Modals On Modals
screen shot 2018-11-26 at 8 10 19 pm

Is this intentional? It feels like something has gone wrong if this is happening.

arielsvn:
## Issue Number

close #444 

Avoid displaying multiple modal dialogs.

## Types of changes

## Functional tests

Modals on download page

## Checklist

### DIFF
--- a/src/components/App/index.js
+++ b/src/components/App/index.js
@@ -21,7 +21,6 @@ import { Provider } from 'react-redux';
 import ErrorBoundary from '../../containers/ErrorBoundary';
 import About from '../About';
 import classnames from 'classnames';
-import { ModalStackProvider } from '../Modal/ModalManager';
 
 /**
  * The 404 route was giving conflicts when used inside App, that's it's extracted into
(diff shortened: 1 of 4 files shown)


Miserlou:
Any chance we can make the modal responsive?

Weird that I have to use non standard scroll buttons to scroll a narrow window on an ultrawide monitor
screen shot 2019-01-09 at 3 06 18 pm

arielsvn:
@Miserlou the modal is responsive, but we have a maximum width for it. Not sure if it will look ok when the modal is wider than the content behind it, we should give it a try.

Would you be ok if we discuss it in a different issue?

arielsvn:
Talked with @dvenprasad about this... The close button on the 2nd level dialog goes to the first dialog, similar to clicking "Back". We need all dialogs to be closed when X is clicked, no matter at what level they were opened. I'll add this change.

PR #501, opened 2019-01-28
Use Formik

### LINKED ISSUE #494: Show error message on Jump to page component when invalid number is entered
Feedback from @dvenprasad on https://github.com/AlexsLemonade/refinebio-frontend/pull/478

We should display an error message when the user enters an invalid page number. Right now, the number just disappears after clicking Go.

image

arielsvn:
## Issue Number

close #494 

Remove redux-form and use Formik instead, since it's easier to use and extend. Also seems that it will integrate better with React Hooks.

## Types of changes

## Functional tests

Checked all forms in the app.

## Checklist

## Screenshots

image

image

### DIFF
--- a/src/components/Pagination/JumpToPageForm.js
+++ b/src/components/Pagination/JumpToPageForm.js
@@ -16,7 +16,7 @@ export default function JumpToPageForm({ onPaginate, totalPages }) {
         page: Yup.number()
           .required('Please enter a valid page number')
           .min(1, 'Page number must be greater than 1')
-          .max(totalPages, 'Page number must be lower than ' + totalPages)
+          .max(totalPages, `Page number must be lower than ${totalPages + 1}`)
       })}
     >
       {({ touched, errors, values }) => (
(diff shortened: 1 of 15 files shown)


src/components/Pagination/JumpToPageForm.js
```diff
@@ -0,0 +1,51 @@
...
+      initialValues={{ page: '' }}
+      onSubmit={({ page }, { resetForm }) => {
+        onPaginate(page);
+        resetForm();
+      }}
+      validationSchema={Yup.object().shape({
+        page: Yup.number()
+          .required('Please enter a valid page number')
+          .min(1, 'Page number must be greater than 1')
+          .max(totalPages, 'Page number must be lower than ' + totalPages)
```
dongbohu: Is ` + totalPages` supposed to be ` + (totalPages + 1)`?

arielsvn: nice catch, fixed it:

image

PR #523, opened 2019-02-09
Big refactor

arielsvn:
## Issue Number

close #309 

## Purpose/Implementation Notes

This includes several changes that I have been wanting to add for a while, among them:

- Simplify `SamplesTable` component. Basically separated the logic of the table form the api fetching. This will make it easier to update the component that we use for the tables once it get's released (https://github.com/tannerlinsley/react-table/issues/1143#issuecomment-459772973)

- Upgraded all node packages to the latest version

- Removed Experiments page from the redux store

- Organize React components in two folders: `components` and `pages`. Components contains generic components such as Buttons, Inputs, Forms, etc and `pages` contains actual pages and the router.

- Rename search page url from `/results` to `/search` to match the API. Also the corresponding components.

- Remove dashboard reducer from redux store, same with the dataset reducers

## Types of changes

* [x] Bugfix (non-breaking change which fixes an issue)
* [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Functional tests

Tested locally.

## Checklist

* [x] Lint and unit tests pass locally with my changes
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] I have added necessary documentation (if appropriate)
* [x] Any dependent changes have been merged and published in downstream modules

## Screenshots

*No visible changes in the app*

PR #526, opened 2019-02-11
Debug circleci integration tests

arielsvn:
I'll squash merge this PR

These commits are to test why the integration tests are failing in #523, it seems the CircleCI configuration we have is not compatible with the latest version of one of the packages we use.


arielsvn:
Will close this, after merging commit 082ab00 into #523 all tests pass.

I'm still unsure what causes the integration tests to fail. Both commits 082ab00 and the latest one here eda38b7 have exactly the same `package.json` file, but the integration tests fail for the latter.

PR #531, opened 2019-02-13
Use fixed layout for filters

### LINKED ISSUE #517: Platform Filter Wonky
It filters, but does a weird flashy thing that wraps off the bottom of the screen:

arielsvn:
## Issue Number

close #517 

Change the layout of filters from sticky to fixed. The can be scrolled independently of the content now.

## Types of changes

## Functional tests

Scroll down in several resolutions.

## Checklist


arielsvn:
Thanks for reviewing @vincerubinetti !!

> Very minor, but maybe change `overflow: 'auto'` to `overflowY: 'auto'`, since it seems like you might have a fixed width for the panel that you wouldn't want to scroll horizontally under any circumstances?

Good catch! will update it!

> Also, I'm assuming the fact that it's fixed now doesn't overlay on the footer of the page in some kind of ugly or obscuring way?

Yes, it doesn't overlay with the footer, but it does uses a fixed value to account for the footer's height

https://github.com/AlexsLemonade/refinebio-frontend/blob/1b4363fbb4014a23e35b0f7f63facc177eecf317/src/pages/Search/ResultFilters/index.js#L73

Not sure if there's a better solution for this.

PR #541, opened 2019-02-20
Detect accession code and show top result

### LINKED ISSUE #519: [Discussion] Don't Sort By Samples If Search Query Is an Accession
I think recent changes to sorting (or perhaps indexing) changed how results in the accession format are returned, ex: https://www.refine.bio/results?q=GSE24528 - I'd expect that accession to be at the top because it's a better match, although it has less samples. So, not really sure what we want to do about this.

arielsvn:
## Issue Number

close #519 

Detects when an user is searching for an accession code, and make an additional API query to retrieve the experiments matching that accession code.

One part that I'm not completely sure is the part to detect what is an accession code. I used this regular expression:

https://github.com/AlexsLemonade/refinebio-frontend/blob/f47365601cbc5d42a1d19df457262faba8ab61e6/src/state/search/actions.js#L46

This will match accession codes with the following:

- `{GSE | ERP | SRP}{3 to 6 digits}`
- `E-{4 upper case letters}-{2 to 4 digits}`

## Types of changes

## Functional tests

Search for accession code

## Checklist

## Screenshots

image

### DIFF
--- a/src/components/SearchInput/index.js
+++ b/src/components/SearchInput/index.js
@@ -22,7 +22,7 @@ export default class SearchInput extends Component {
       <Formik
         ref={this.formik}
         initialValues={{ search: this.props.searchTerm }}
-        onSubmit={({ search }) => this.props.onSubmit(search)}
+        onSubmit={({ search }) => this.props.onSubmit(search.trim())}
       >
         {() => (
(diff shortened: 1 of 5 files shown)


Miserlou:
Love it!

Issue #543, opened 2019-02-21
Application Size Is Over 1MB


Miserlou:
![screen shot 2019-02-21 at 10 29 48 am](https://user-images.githubusercontent.com/139987/53180515-b3073680-35c3-11e9-9ed9-f476005cf8b0.png)

Maybe I'm being an Old Person^tm, but this seems incredibly large for a simple web application. Why is this so large? Is it not minified? It does use the cached version on the next load, so that's good.

Furthermore, it looks like the size and the transfered sized are the same, meaning that we're not using `gzip` (or similar) for our frontend assets. 

It also looks like we're not loading our external scripts async since hotjar blocked the page render for me, but that's probably a different ticket.

arielsvn:
This might be a good opportunity to also add some of the recommendations from [google's pagespeed insights](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.refine.bio%2Fsearch).

PR #547, opened 2019-02-22
Remove icon library and use react-icons

arielsvn:
## Issue Number

#543 

Removes all usages of [Ionicons](https://ionicons.com/v1/), we needed only <10 icons from them and we were loading the entire font.

Ionicons was replaced with [react-icons](https://github.com/react-icons/react-icons), the good thing about it is that only the used icons will end in the JS file as svg.

According to pagespeed this should reduce another ~1sec from the render time.

## Types of changes

## Functional tests

All icons replaced, ensure they look the same.

## Checklist


arielsvn:
Forgot to mention this also includes a fix to ensure all urls in the sitemap are valid.

src/components/Spinner.js
```diff
@@ -5,7 +5,7 @@ import spinner from '../common/images/spinner.gif';
 export default function Spinner() {
   return (
     <div className={styles.spinner}>
-      <img src={spinner} width="59" height="54"/>
+      <img src={spinner} width="59" height="54" alt="loading" />
```
Miserlou: inline style attributes >:|

arielsvn: Seems that this [is recommended](https://stackoverflow.com/a/6784879/763705), browsers use these attributes as "hints for the rendering".

arielsvn: took me a while to find that

PR #557, opened 2019-03-07
Redirect to `/search` from `/ results`

### LINKED ISSUE #554: refine.bio google search results link to 404 pages :(
Good news: we are the first result! 

screen shot 2019-03-05 at 4 48 43 pm

Bad news: But it links to our 404 page. :(

arielsvn:
## Issue Number

close #554 

We changed the search page from `/results` to `/search` but google has some of the old urls indexed. This ensures we redirect appropriately.

## Types of changes

## Functional tests

Visit `/results?q=human`

## Checklist

### DIFF
--- a/src/pages/App.js
+++ b/src/pages/App.js
@@ -2,7 +2,7 @@ import React from 'react';
 import Helmet from 'react-helmet';
 import { Provider } from 'react-redux';
 import classnames from 'classnames';
-import { Router, Route, Switch } from 'react-router-dom';
+import { Router, Route, Switch, Redirect } from 'react-router-dom';
 
 import history from '../history';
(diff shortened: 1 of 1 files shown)


Miserlou:
What type of HTTP code does this return? We want this to be a permanent redirect.

arielsvn:
It's always 200, since this is a SPA and the routing happens in the frontend after the page has been loaded. I was hoping google to drop these urls after the sitemap get's uploaded.

Miserlou:
Ah. Hm, I don't know anything about how to handle legacy redirects inside an SPA. I guess this works but we'll need to circle back around to make sure things have been updated properly in about a month.

Issue #638, opened 2019-05-21
Search doesn't always default to best match results


dvenprasad:
Got this comment via hotjar
>I was trying to update my search from within the search results window.
I started by querying "neural stem cells", then added "embryonic" inside the results menu. When I updated the search, the results were no longer relevant to NSCs or embryonic.

The results are there but when the user modifies/starts a new search from the results page, the new results defaults to `Most number of samples` instead of `Best Match`

I tried it with different terms and it still defaulted to `Most number of samples`.

arielsvn:
fixed by #644

PR #644, opened 2019-05-28
new search defaults to best match

davidsmejia:
## Issue Number
#638 

When trigger search is called (when a new search is submitted from search results page) the ordering is set to Ordering.MostSamples. This changes that to use Ordering.BestResults instead. 

* [ x] Bugfix (non-breaking change which fixes an issue)

## Functional tests

This is untested and I haven't set up anything locally.

## Checklist

## Screenshots

None

### DIFF
--- a/src/state/search/actions.js
+++ b/src/state/search/actions.js
@@ -167,7 +167,7 @@ export const triggerSearch = searchTerm => (dispatch, getState) => {
       query: searchTerm,
       page: 1,
       filters: {},
-      ordering: Ordering.MostSamples,
+      ordering: Ordering.BestMatch,
     })
   );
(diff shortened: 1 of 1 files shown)


davidsmejia:
Note: If you search something -> update the ordering -> modify your query. The ordering will still revert to Ordering.BestMatch instead of whichever ordering you had selected before submitting a new query.

arielsvn:
> If you search something -> update the ordering -> modify your query. The ordering will still revert to Ordering.BestMatch instead of whichever ordering you had selected before submitting a new query.

I think this should be the expected behavior, since we also reset the filters, page numbers, etc

Issue #648, opened 2019-05-30
`useLoader` causes infinite calls to the server


arielsvn:
Probably because it was not implemented correctly. It would be good to add some tests to it to ensure this doesn't happen again.

These articles seem to explain the issue in detail.

https://www.robinwieruch.de/react-hooks-fetch-data/
https://overreacted.io/a-complete-guide-to-useeffect/

Added https://github.com/AlexsLemonade/refinebio-frontend/pull/646 as a temporary fix

arielsvn:
fixed in #651 and #658

PR #649, opened 2019-05-31
wrap useEffect callback in React.useCallback

davidsmejia:
## Issue Number
#648

This is an attempt to make the linter not helicopter parent with the added benefit of not sending an API request every time the screen refreshes.

This has the same functionality as #646 without manually silencing the linter. After reading the docs I honestly can't tell if this is how it's meant to be implemented or just a sanctioned workaround. 

Essentially the changes are:
1) React.useEffect calls refreshDataCallback with updateProps as a dependency
2) Adds new function called refreshDataCallback that encapsulates the incrementing of 'mounted.current' and fetchDataCallback's invocation into its own React.useCallback to be used when manually refreshing or refreshing from a change in one of passed in 'updateProps'. 

## Types of changes

## Functional tests
  The setInterval refresh() and the range select on the dashboard look to be functional.
  I also tried to abuse the range select but the most recent range fetched was always the one rendered.
  
  Tested Here:
    -src/pages/Dashboard/index.js
  Still needs to be tested here:
    -src/pages/ExecDashboard/index.js
    -src/pages/ExecDashboard/SamplesProcessedBlock.js
    -src/pages/SpeciesCompendia/DownloadCompendia.js

## Checklist

## Screenshots
Sadly, no.

### DIFF
--- a/src/components/Loader/index.js
+++ b/src/components/Loader/index.js
@@ -110,20 +110,19 @@ export function useLoader(fetch, updateProps = []) {
     [fetch, state]
   );
 
-  React.useEffect(
-    () => {
+  const refreshDataCallback = React.useCallback(
+    function incrementAndFetchData() {
       mounted.current += 1;
       fetchDataCallback(mounted.current);
     },
-    updateProps // eslint-disable-line react-hooks/exhaustive-deps
+    [fetchDataCallback]
   );
+
+  React.useEffect(refreshDataCallback, updateProps);
 
   return {
     ...state,
     hasError: !!state.error,
-    refresh: async () => {
-      mounted.current += 1;
-      await fetchDataCallback(mounted.current);
-    },
+    refresh: refreshDataCallback,
   };
 }


src/components/Loader/index.js
```diff
@@ -110,20 +110,19 @@ export function useLoader(fetch, updateProps = []) {
...
+  const refreshDataCallback = React.useCallback(
+    function incrementAndFetchData() {
       mounted.current += 1;
       fetchDataCallback(mounted.current);
     },
-    updateProps // eslint-disable-line react-hooks/exhaustive-deps
+    [fetchDataCallback]
   );
 
+  React.useEffect(refreshDataCallback, updateProps);
```
arielsvn: Very interesting solution! Thanks for contributing it! I ran [several tests](https://github.com/AlexsLemonade/refinebio-frontend/blob/320c708730cee1d0e353f2b3f200dc00396fdbae/src/components/Loader/loader.test.js) and this passed all of them.

It seems eslint complains if you use this syntax:

```js
React.useEffect(() => refreshDataCallback(), updateProps);
```

Why do you think this happens? I'm wondering if the original version has the same issues but is a case that the rule can't detect.

arielsvn:
Made another try at #651, I'd be very interested in hearing your thoughts about it.

src/components/Loader/index.js
davidsmejia: Hey, I think the fact that it throws a warning makes sense... however the specific warning that 'updateProps' is not an array... I haven't made any sense of that.

arielsvn: I get two warnings:

> [eslint] React Hook React.useEffect was passed a dependency list that is not an array literal. This means we can't statically verify whether you've passed the correct dependencies. (react-hooks/exhaustive-deps)

I don't think there's any way around this one, since the idea is to have the effect run whenever the values of `updateProps` change, and that must be a parameter :/

The one that worried me more was:

> [eslint] React Hook React.useEffect has a missing dependency: 'refreshDataCallback'. Either include it or remove the dependency array. (react-hooks/exhaustive-deps)

Seems that the recommendation is to always include [functions in the list of effect dependencies](https://overreacted.io/a-complete-guide-to-useeffect/#moving-functions-inside-effects).

I think we should merge this one since it's better than the current solution in `dev`, and see if #651 makes any difference.

PR #651, opened 2019-05-31
Arielsvn/648 useloader fix

arielsvn:
## Issue Number

#648 

Another attempt to fixing `useLoader`. [Found a way](https://overreacted.io/a-complete-guide-to-useeffect/#why-usereducer-is-the-cheat-mode-of-hooks) to remove `state` as a dependency from the effect that called `fetch` so that we can avoid the infinite cycles that caused #648 

## Types of changes

## Functional tests

Added tests for this.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "test:integration": "jest --runInBand -c integration/jest.config.js"
   },
   "devDependencies": {
+    "@testing-library/react": "^8.0.1",
     "axios": "^0.18.0",
     "eslint": "^5.16.0",
     "eslint-config-airbnb": "^17.1.0",
@@ -69,6 +70,7 @@
     "eslint-plugin-react-hooks": "^1.6.0",
     "flow-bin": "^0.68.0",
     "husky": "^1.3.1",
+    "jest-dom": "^3.4.0",
     "jest-puppeteer": "^3.7.0",
     "lint-staged": "^7.0.0",
     "node-sass": "^4.11.0",
--- a/src/common/helpers.test.js
+++ b/src/common/helpers.test.js
@@ -61,7 +61,7 @@ describe('getQueryParamObject', () => {
 });
 
 describe('Ajax', () => {
-  beforeEach(function() {
+  beforeEach(() => {
     global.fetch = jest
       .fn()
       .mockImplementation(() => Promise.resolve({ ok: true, json: () => {} }));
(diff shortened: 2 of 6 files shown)


davidsmejia:
I think this is the right idea. It's way cleaner anyway. If you want to stop the infinite loops (you probably do) you can add a React.useCallback on Dashboard/index.js but, IMO it's not the prettiest implementation.

https://github.com/davidsmejia/refinebio-frontend/commit/1a529c33fb0bdf32a24f14b9bf14ef516753feff

arielsvn:
> If you want to stop the infinite loops (you probably do) you can add a React.useCallback on Dashboard/index.js but, IMO it's not the prettiest implementation.

Good catch! I'll try to debug this tomorrow to see why it happened. Thanks for taking a look!

In https://github.com/davidsmejia/refinebio-frontend/commit/1a529c33fb0bdf32a24f14b9bf14ef516753feff one test did not pass.

davidsmejia:
Hey that test fails because in https://github.com/davidsmejia/refinebio-frontend/commit/1a529c33fb0bdf32a24f14b9bf14ef516753feff because in this implementation of useLoader the dependency is the callback itself not the [param] that is redefined it is the function passed to the hook (which is why if we add a useCallback hook it will keep the same function ref and not be redefined) - this is because the Dashboard is reinitialized when that value is changed.

It looks like /dashboard renders 4 times on load before user interaction, 3 on initial load and again when the useLoader gets its ajax call returned, meanwhile, /search loads 3 initial times as well. I'm looking into that now.

arielsvn:
Added `React.useCallback` in the dashboard as suggested by @davidsmejia, that seems to fix the infinite loop in the dashboard.

davidsmejia:
For what it's worth, I think this looks great. Silencing the eslinter seems appropriate too as there is no way to statically apply the dependencies in the useLoader scope.

As for the multiple render calls also seem unavoidable as they are a consequence of the useLoader setting state causing React to do a diff it's not actually rendering anything to screen.

arielsvn:
>  Silencing the eslinter seems appropriate too as there is no way to statically apply the dependencies in the useLoader scope.

Good call, just made the change. Thanks again for keeping an eye on this! Let's get it merged.

@kurtwheeler this should be good to go, can you help us with an 👍 ?

PR #657, opened 2019-06-05
refactor useLoader to reduce loops

davidsmejia:
## Issue Number

#648 

I was working on a house cleaning PR implementing some of the new esLint rules when I stumbled across the executive-dashboard doing infinite loops. 👎 So this is a fix for that.

The idea here is that the useLoader should only return it's own state. That state should only be set in the dataFetchReducer. The fetched data should only be called when the watched property changes. Also it should fetch data initially on the first render/page load.

Note:
Occasionally in development mode you might see a console warning if the dashboard was unmounted and the fetch resolves or if an invalidated fetch resolves it would set state and React would warn about a potential memory leak. I believe In the next version of React (16.9.x) there will be better ways of implementing this functionality. Hypothetically, if the days were longer you could implement an AbortController in the return function (the way you cleanup on unmount) of the useEffect inside useLoader. But refactoring that code to make this air tight would take on more risk than reward IMO.

This PR's changes are in two parts.

1. A refactor of useLoader.
 - adds hasError: true/false to the state changes in dataFetchReducer
 - Defines fetchDataCallback at the top of useLoader so it can be set to state
 - adds ...updateProps to the dependencies of the fetchDataCallback and silences eslint warnings
 - removes fetchDataCallback from dependencies of the useEffect
 - returns only the state object

2. Simplifies the pages/dashboard/index
 - adds isLoading to deconstructed useLoader return
 - removes references to updateCharts in favor of isLoading
 - removes the useCallback wrapped fetchCallback due to useLoader refactor 

## Types of changes

## Functional tests

No infinite loops on dashboard or executive dashboard. 
Tested these events:
- dashboard load
- dashboard range select callback
- dashboard refresh from interval
- executive dashboard
- executive dashboard refresh from interval
- executive dashboard SamplesProcessedBlock load
- executive dashboard SamplesProcessedBlock range select callback

## Checklist

### DIFF
--- a/src/components/Loader/index.js
+++ b/src/components/Loader/index.js
@@ -114,19 +114,9 @@ const dataFetchReducer = (state, action) => {
  * @param {*} updateProps Similar to `useEffect` dependencies, you can add any values here that invalidate the result of `fetch`
  */
 export function useLoader(fetch, updateProps = []) {
-  // Using a reducer helps remove the `state` dependency from the effect below
-  // React guarantees that `dispatch` is unique accross renders.
-  // https://overreacted.io/a-complete-guide-to-useeffect/#why-usereducer-is-the-cheat-mode-of-hooks
-  const [state, dispatch] = React.useReducer(dataFetchReducer, {
-    error: null,
-    isLoading: true,
-    data: null,
-  });
-
   const fetchDataCallback = React.useCallback(
     async function fetchData() {
       dispatch({ type: 'FETCH_INIT' });
-
       try {
         const data = await fetch();
         dispatch({ type: 'FETCH_SUCCESS', payload: data });
@@ -134,16 +124,23 @@ export function useLoader(fetch, updateProps = []) {
         dispatch({ type: 'FETCH_FAILURE', payload: error });
       }
     },
-    [fetch]
+    [fetch, ...updateProps] // eslint-disable-line react-hooks/exhaustive-deps
   );
+  // Using a reducer helps remove the `state` dependency from the effect below
+  // React guarantees that `dispatch` is unique accross renders.
+  // https://overreacted.io/a-complete-guide-to-useeffect/#why-usereducer-is-the-cheat-mode-of-hooks
+  const [state, dispatch] = React.useReducer(dataFetchReducer, {
+    refresh: fetchDataCallback,
+    isLoading: true,
+    error: null,
+    hasError: false,
+    data: null,
+  });
 
+  // Call fetch when properties change.
   React.useEffect(() => {
     fetchDataCallback();
-  }, [fetchDataCallback, ...updateProps]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [...updateProps]); // eslint-disable-line react-hooks/exhaustive-deps
 
-  return {
-    ...state,
-    hasError: !!state.error,
-    refresh: fetchDataCallback,
-  };
+  return state;
 }
--- a/src/pages/Dashboard/index.js
+++ b/src/pages/Dashboard/index.js
@@ -11,18 +11,15 @@ import { getQueryParamObject, formatBytes } from '../../common/helpers';
 import './Dashboard.scss';
 
 function Dashboard(props) {
-  const [chartUpdating, setChartUpdating] = React.useState(true);
   let { range: rangeParam } = getQueryParamObject(props.location.search);
   if (!['day', 'week', 'month', 'year'].includes(rangeParam)) {
     rangeParam = 'day';
   }
   const [range, setRange] = React.useState(rangeParam);
-  const fetchCallback = React.useCallback(async () => {
+  const { data, refresh, isLoading, hasError } = useLoader(async () => {
     const stats = await fetchDashboardData(range);
-    setChartUpdating(false);
     return getDashboardChartConfig(stats, range);
   }, [range]);
-  const { data, refresh, hasError } = useLoader(fetchCallback, [range]);
 
   // refresh data every 10 mins
   useInterval(() => {
@@ -44,15 +41,12 @@ function Dashboard(props) {
             { label: 'Last Month', value: 'month' },
             { label: 'Last Year', value: 'year' },
           ]}
-          onChange={range => {
-            setChartUpdating(true);
-            setRange(range);
-          }}
+          onChange={selected => setRange(selected)}
         />
 
         <p>* All dates in UTC</p>
 
-        {!data || chartUpdating ? (
+        {!data || isLoading ? (
           <Spinner />
         ) : (
           data.map(section => {


src/components/Loader/index.js
```diff
@@ -114,36 +114,33 @@ const dataFetchReducer = (state, action) => {
...
     fetchDataCallback();
-  }, [fetchDataCallback, ...updateProps]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [...updateProps]); // eslint-disable-line react-hooks/exhaustive-deps
 
-  return {
-    ...state,
-    hasError: !!state.error,
-    refresh: fetchDataCallback,
-  };
+  return state;
```
arielsvn: Any advantage to returning `state` here? seems unnecessary `hasError` and `refresh` as part of the state

src/pages/Dashboard/index.js
```diff
@@ -44,15 +41,12 @@ function Dashboard(props) {
...
-            setChartUpdating(true);
-            setRange(range);
-          }}
+          onChange={selected => setRange(selected)}
         />
 
         <p>* All dates in UTC</p>
 
-        {!data || chartUpdating ? (
+        {!data || isLoading ? (
```
arielsvn: good call!

src/components/Loader/index.js
```diff
@@ -114,36 +114,33 @@ const dataFetchReducer = (state, action) => {
...
+    error: null,
+    hasError: false,
+    data: null,
+  });
 
+  // Call fetch when properties change.
   React.useEffect(() => {
     fetchDataCallback();
-  }, [fetchDataCallback, ...updateProps]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [...updateProps]); // eslint-disable-line react-hooks/exhaustive-deps
```
arielsvn: hmm seems that there's a [strong recommendation](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) to list functions as dependencies in effects.

src/components/Loader/index.js
```diff
@@ -114,36 +114,33 @@ const dataFetchReducer = (state, action) => {
...
-
       try {
         const data = await fetch();
         dispatch({ type: 'FETCH_SUCCESS', payload: data });
       } catch (error) {
         dispatch({ type: 'FETCH_FAILURE', payload: error });
       }
     },
-    [fetch]
+    [fetch, ...updateProps] // eslint-disable-line react-hooks/exhaustive-deps
```
arielsvn: why add `updateProps` here? it doesn't seem to be a dependency of this callback

Any disadvantages with having `React.useCallback` in the dashboard?

src/components/Loader/index.js
davidsmejia: Yes, you are correct. This is has no impact on functionality. However, I thought the advantage of declaring ```hasError``` and ```refresh``` as part of the default state was improved readability. There would be no side effects if this was undone.

src/pages/Dashboard/index.js
davidsmejia: Thanks 👍

src/components/Loader/index.js
arielsvn: I see, I was a bit confused with the callback since the state references it and the callback uses `dispatch` defined below. Maybe we can leave `hasError` in the state and move `refresh` out? 

With redux, they [recommend not having functions in the state](https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state). Not sure if there are any drawbacks with having them in `useReducer` hook?

davidsmejia:
@arielsvn what are your thoughts on pulling out the useEffect in useLoader and returning a wrapped function for requesting data/refreshes?

https://github.com/davidsmejia/refinebio-frontend/compare/dev...davidsmejia/support/useLoader-pull-out-useEffect

arielsvn:
hmm but then the hook won't do the initial fetch when the component is mounted right? IMO that's an important part of it since it will simplify usages in simpler components.

Did you see [this blog post](https://www.robinwieruch.de/react-hooks-fetch-data/)? the current version of `useLoader` is highly inspired by it.

davidsmejia:
That's right. If that was implemented it wouldn't fetch on initial render unless you put it in a ```useEffect``` with an array literal of your watched properties aka dependencies aka updateProps.. Which is more powerful and less error prone IMO at the cost of being less DRY. I thought you'd prefer it to wrapping the ```async fetch``` in a ```React.useCallback```... 

However, if keeping the initial load behavior is a priority an alternative solution with the added benefit of not having to wrap your ```async fetch``` in a ```useCallback``` would be:

```javascript
export function useLoader(fetch, updateProps = []) {
  // memoize the fetch
  const fetchRef = React.useRef(fetch);
  React.useEffect(() => {
    fetchRef.current = fetch;
  }, [fetchRef, fetch]); // updates more often than needed but its never stale and doesnt affect rendering because its in a ref
  // memoize the updateProps
  const updatePropsRef = React.useRef(updateProps);
  // update the memo if not deep equal
  if (!isEqual(updatePropsRef.current, updateProps)) {
    updatePropsRef.current = updateProps;
  }
  // pull out the memo from the ref to use in the side effect dependencies
  const {current: updatePropsMemo} = updatePropsRef;
  // Using a reducer helps remove the `state` dependency from the effect below
  // React guarantees that `dispatch` is unique accross renders.
  // https://overreacted.io/a-complete-guide-to-useeffect/#why-usereducer-is-the-cheat-mode-of-hooks
  const [state, dispatch] = React.useReducer(dataFetchReducer, {
    error: null,
    isLoading: true,
    data: null,
  });

  const fetchDataCallback = React.useCallback(async () => {
    dispatch({ type: 'FETCH_INIT' });
    try {
      const data = await fetchRef.current();
      dispatch({ type: 'FETCH_SUCCESS', payload: data });
    } catch (error) {
      dispatch({ type: 'FETCH_FAILURE', payload: error });
    }
  }, [fetchRef]); // never changes
  // useLoader will fetch Data when updateProps changes
  React.useEffect(() => {
    fetchDataCallback();
  }, [fetchDataCallback, updatePropsMemo]);
  // returned for use in your component
  return {
    ...state,
    hasError: !!state.error,
    refresh: fetchDataCallback,
  };
}

arielsvn:
I like that solution, nice usage of `useRef` to keep a reference to the latest `fetch` function. I ran the tests against it and they all passed.

It would be cool to get that merged, but I have a question, why have the `updatePropsRef.current` instead of having `...updateProps` as dependencies directly?

davidsmejia:
OK sending a PR now will reference this in the notes. 

To answer your question - that was for future proofing the useLoader dependency list since it would not trigger a refetch if you passed in an array whose contents were changed since useEffect's comparator is Object.is. I won't include it in the PR since it doesn't affect any current implementations. Just thought it would make it more robust. I'm open to revisiting it later if you'd like.

Issue #627, opened 2019-05-03
Use JavaScript style guide


arielsvn:
Just noticed [React](https://reactjs.org/docs/how-to-contribute.html#style-guide) uses [Airbnb's style guide](https://github.com/airbnb/javascript). We should start enforcing those rules as well.

Reading through the guide, we meet most of the conditions but not all.

arielsvn:
#640 Adds most of the rules, although some are not marked to trigger errors because the fixes will require bigger refactors. I'll fix them gradually.

```
    'react/no-did-update-set-state': 'warn',
    'dot-notation': 0,
    'no-underscore-dangle': 'warn',
    'react/sort-comp': 'warn',
    'no-shadow': 'warn',
    'prefer-destructuring': 'warn',
    'no-param-reassign': 'warn',
```

PR #666, opened 2019-06-18
eslint fixes for accordion.js

davidsmejia:
## Issue Number

#627 

Makes components/accordion.js compliant with eslint.
- Moves render to the bottom
- renames ```_totalChildren``` to ```getChildrenCount```
- moves```getChildrensCount``` helper to the top
- Renames private ```_toggleAll``` to ```toggleAll```
- Renames private ```_toggleElement``` to ```toggleElement```

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -14,6 +14,30 @@ export class Accordion extends React.Component {
     };
   }
 
+  _toggleAll() {
+    if (!this.state.activeElements.includes(false)) {
+      this.setState({
+        activeElements: new Array(this._totalChildren()).fill(false),
+      });
+    } else {
+      this.setState({
+        activeElements: new Array(this._totalChildren()).fill(true),
+      });
+    }
+  }
+
+  _toggleElement(index) {
+    this.setState(state => ({
+      activeElements: state.activeElements.map((x, i) =>
+        i === index ? !x : x
+      ),
+    }));
+  }
+
+  _totalChildren() {
+    return React.Children.count(this.props.children);
+  }
+
   render() {
     const children = React.Children.map(
       this.props.children,
@@ -41,30 +65,6 @@ export class Accordion extends React.Component {
         {children}
       </div>
     );
-  }
-
-  _toggleAll() {
-    if (!this.state.activeElements.includes(false)) {
-      this.setState({
-        activeElements: new Array(this._totalChildren()).fill(false),
-      });
-    } else {
-      this.setState({
-        activeElements: new Array(this._totalChildren()).fill(true),
-      });
-    }
-  }
-
-  _toggleElement(index) {
-    this.setState(state => ({
-      activeElements: state.activeElements.map((x, i) =>
-        i === index ? !x : x
-      ),
-    }));
-  }
-
-  _totalChildren() {
-    return React.Children.count(this.props.children);
   }
 }
 
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -18,7 +18,7 @@ export class Accordion extends React.Component {
     return React.Children.count(this.props.children);
   }
 
-  _toggleAll() {
+  toggleAll() {
     if (!this.state.activeElements.includes(false)) {
       this.setState({
         activeElements: new Array(this.getChildrenCount()).fill(false),
@@ -30,7 +30,7 @@ export class Accordion extends React.Component {
     }
   }
 
-  _toggleElement(index) {
+  toggleElement(index) {
     this.setState(state => ({
       activeElements: state.activeElements.map((x, i) =>
         i === index ? !x : x
@@ -45,7 +45,7 @@ export class Accordion extends React.Component {
         child &&
         React.cloneElement(child, {
           isExpanded: this.state.activeElements[index],
-          onToggle: () => this._toggleElement(index),
+          onToggle: () => this.toggleElement(index),
         })
     );
     return (
@@ -53,7 +53,7 @@ export class Accordion extends React.Component {
         {!this.props.hideExpandAll && children.length > 2 && (
           <Checkbox
             className="accordion__expand-all"
-            onClick={() => this._toggleAll()}
+            onClick={() => this.toggleAll()}
             checked={!this.state.activeElements.includes(false)}
             name="expand-all"
             readOnly
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -14,6 +14,10 @@ export class Accordion extends React.Component {
     };
   }
 
+  getChildrenCount() {
+    return React.Children.count(this.props.children);
+  }
+
   _toggleAll() {
     if (!this.state.activeElements.includes(false)) {
       this.setState({
@@ -32,10 +36,6 @@ export class Accordion extends React.Component {
         i === index ? !x : x
       ),
     }));
-  }
-
-  getChildrenCount() {
-    return React.Children.count(this.props.children);
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


davidsmejia:
@arielsvn Since the linter doesn't like the underscore convention for "private" methods and variables would you like to move it outside of the export instead of just renaming it? It might be unnecessary though because it's rare that you would call a method on a component directly anyway.

davidsmejia:
@arielsvn If you have a preferences or requests for how I approach any of this I'm open to hearing it. I started by doing it by each eslint rule but I think thats just too much to ask someone to review. I wanted to make the PR's smaller and easier to validate the changes.

Also was just curious why some of these components are top level in components/ would it make sense to refactor these into folders and update the imports as I go along?

arielsvn:
> Since the linter doesn't like the underscore convention for "private" methods and variables would you like to move it outside of the export instead of just renaming it?

What do you mean? Make `Accordion` a function component?

> If you have a preferences or requests for how I approach any of this I'm open to hearing it. I started by doing it by each eslint rule but I think that's just too much to ask someone to review.

Yes, I think that's probably the best approach, since they will be easier to review. 

I should mention, in the case of the eslint rules, I'm not sure that all of the ones marked as warnings should be enforced. I'd be open to hear your thoughts if you think some of them should just be disabled (or fully enabled). For example, [camelcase](https://github.com/airbnb/javascript#naming--camelCase) is tricky to implement since the fields from the API are underscore case.

> Also was just curious why some of these components are top level in components/ would it make sense to refactor these into folders and update the imports as I go along?

We had a [discussion about this](https://github.com/AlexsLemonade/refinebio-frontend/pull/361#discussion_r222701918) a while ago, personally I don't like when a lot of file names are `index.js`, so for simple components I started adding them as top level components, as long as they only were two files (js + scss). Hopefully this won't cause any issues, [ facebook has >30k top level components](https://www.reddit.com/r/reactjs/comments/6al7h2/facebook_has_30000_react_components_how_do_you/dhgruqh/)

davidsmejia:
> What do you mean? Make Accordion a function component?

Well I was suggesting just pulling the definition outside of the exported scope. So it was defined in the module. However you might prefer this:
```javascript
const toggleAll = new Symbol('toggleAll');
export class Accordion extends React.Component { 

[...]

[toggleAll]() {
    if (!this.state.activeElements.includes(false)) {
      this.setState({
        activeElements: new Array(this.getChildrenCount()).fill(false),
      });
    } else {
      this.setState({
        activeElements: new Array(this.getChildrenCount()).fill(true),
      });
    }
  }

[...] 

}
```
 Then you wouldn't be able to access the method outside of the class. 

> For example, camelcase is tricky to implement since the fields from the API are underscore case.

In almost all cases you can rename them when you use them and this will satisfy the linter.

```javascript
export async function getGenomeBuild(organism) {
  try {
    const { assembly_name: assemblyName } = await Ajax.get('/transcriptome_indices/', {
      organism,
      length: 'long',
    });

    return assemblyName;
  } catch (e) {
    return null;
  }
}
```

> We had a discussion about this a while ago, personally I don't like when a lot of file names are index.js,

That sounds good. I typically ```⌘ + T``` to open a file in Atom anyway. I don't think either/or would be an issue. Though, if it's ever opened to a vote put me down for making them all top level and just renaming the images by prepending ```ComponentName-a-thing.svg```  to indicate ownership. It would improve scannability in my editor's file tree because it sorts folders in alphabetical order then files in alphabetical order. Plus the open tabs would have more meaningful content instead of 14 ```index.js``` files open.

arielsvn:
> Well I was suggesting just pulling the definition outside of the exported scope. So it was defined in the module. However you might prefer this:

Oh first time I see that sintax: `toggleAll = new Symbol('toggleAll')` and then `[toggleAll]() {...}`, seems like it would be too much trouble just for a private method. However, just from curiosity, how would you call that method? `this[toggleAll]()` ?

> In almost all cases you can rename them when you use them and this will satisfy the linter.

Yeah, problem is that we would have to do the same when sending data to the API as well.

> if it's ever opened to a vote put me down for making them all top level and just renaming the images by prepending ComponentName-a-thing.svg to indicate ownership.

I agree to this, although we can judge on a case by case basis. For example [SamplesTable](https://github.com/AlexsLemonade/refinebio-frontend/tree/dev/src/components/SamplesTable) has multiple sub-components that are only used by `SamplesTable`, in this case a folder might be better suited. Do you agree?

Most of the other ones that we have in the `/components` folder can be placed in the top level as you said.

davidsmejia:
> Oh first time I see that sintax: toggleAll = new Symbol('toggleAll') and then [toggleAll]() {...}, seems like it would be too much trouble just for a private method. However, just from curiosity, how would you call that method? this[toggleAll]() ?

thats correct except I accidentally put ```new Symbol()``` should be 
I'll correct the error for posterity. 

```javascript
const toggleAll = Symbol('toggleAll');
export ...
```

Since you don't have access to the symbol outside of the scope of the module the method would be inaccessible from the outside. I agree this is too much for this use case. But a neat feature of the language nonetheless.

> Yeah, problem is that we would have to do the same when sending data to the API as well.

Yeah I hear you it does seem silly to change every snake case to camelcase and the back again for API consumption. Like the following which the linter is ok with. (this is an example and probably only needs to be implemented in something that returns directly from a ```wait Ajax.verb``` but it illustrates the point.

```javascript
/**
 * edit some parameter of the dataset
 * @param {*} params additional params to be sent to the API
 */
export const editDataSet = ({ dataSetId, ...params }) => async (
  dispatch,
  getState
) => {
  const { dataSet } = getState().download;
  const {
    data,
    is_processing: isProcessing,
    is_processed: isProcessed,
    aggregate_by: aggregateBy,
    scale_by: scaleBy,
    quantile_normalize: quantileNormalize,
  } = await Ajax.put(`/dataset/${dataSetId}/`, {
    data: dataSet,
    ...params,
  });

  dispatch(
    updateDownloadDataSet({
      dataSet: data,
      is_processing: isProcessing,
      is_processed: isProcessed,
      aggregate_by: aggregateBy,
      scale_by: scaleBy,
      quantile_normalize: quantileNormalize,
    })
  );
};

```

davidsmejia:
> I agree to this, although we can judge on a case by case basis. For example SamplesTable has multiple sub-components that are only used by SamplesTable, in this case a folder might be better suited. Do you agree?

Interesting interesting. 

I guess i'm just a fan of the consistency of either direction. But i hear what you're saying, maybe it would be appropriate to have ```SamplesTable.js``` top-level and sub-components in ```./components/SamplesTable/SampleDetalsLoader.js```
```javascript 
import SamplesTable from '../../components/SamplesTable';
````
Also you could modify your webpack resolver to look for ```folder/folder.js``` instead of ```folder/index.js``` and keep the existing ```import folder from 'folder'``` but i like the previous example better since it would mean only sub-components are in folders and no index.js files would be needed and no potentially version-upgrade-breakable modifications to webpack config introduced

thoughts?

arielsvn:
> actually it might be nice to tell webpack to look resolve from the components folder anyway so you could just put `import componentName from 'componentName'`

Is there any way to implement this without ejecting the webpack configuration? right now all our webpack config is managed by create-react-app. My biggest concern with this approach is that it might break the IDE, since it won't know from where the component is being imported.

Regarding the imports, I'd be happy if something like the following worked: `import {Button, Input} from 'components'`. I think I saw some project where they had an `index.js` file under `/components` that exported all components, but I can't remember where now. Another approach I saw recently could be having a `node_modules` folder in `/src` with the components ([ref](https://twitter.com/Rich_Harris/status/1139908960520220677?s=20)).

Issue #614, opened 2019-04-23
The horizontal scroll button "screams" too much


dvenprasad:
Limit the animation of the horizontal scroll until the user has scrolled horizontally. Don't repeat it after that.

arielsvn:
closed by https://github.com/AlexsLemonade/refinebio-frontend/pull/679

PR #668, opened 2019-06-20
Reduce Button Shake for Horizontal Scroll

davidsmejia:
## Issue Number

#614 

Reduces the button shake animation to run only when it's initially displayed in viewport.

## Types of changes

## Functional tests

Animates when:
- user scrolls down and button is in viewport
- user resizes screen to reveal right button

Does Not animate when:
- resizing more than once
- scrolling page up and down while button is enabled
- scrolling table to the right then the left again (revealing the button again)

## Checklist

## Screenshots

None. If you would like to see a gif I can put one together.


arielsvn:
Cool, nice use of the hooks to solve this! 🔥 

The experiments page seems to be fine, but the shaking still happens on the dataset page :/ where the button component is inside a modal dialog an it gets recreated every time the dialog is displayed.

davidsmejia:
Good catch. Is there a way to run a single processor job locally by ```assession_code```?

arielsvn:
Not sure I understand, what do you mean?

davidsmejia:
There is a good chance I don't know what I mean.

I'm trying to add a sample to my dataset via the front end but I don't have a processed sample to add locally. So, I was thinking it would make sense to process a sample but I'm unsure how to proceed.

davidsmejia:
To clarify I'd like to be able to add this to my dataset.
Screen Shot 2019-06-20 at 1 45 12 PM

arielsvn:
Oh I see, yes running a [local copy of the API is tricky](https://github.com/AlexsLemonade/refinebio#running-locally). You could also run it against our prod API with: 

`REACT_APP_API_HOST=https://api.refine.bio yarn start`

We have a script `run.sh` at the top level for convenience :) Does that work?

davidsmejia:
Yup that's perfect thanks!

davidsmejia:
@arielsvn I pushed a solution to the dataset page - it will keep the animation to once per load of the component... meaning it wont fire again if you navigate through the app to another instance of a table using the horizontal scroll since it would have already been loaded and the animation was shown to you at some point during your interaction with the app... if you want it to fire on every new page event that is doable too... somehow...

But If I'm being honest I think I might be trying to solve the wrong problem. What are your thoughts on having both buttons fade in whenever visible instead. Making the animation less attention greedy and keeping it consistent IMO would make for a better user experience than whats currently live and would be a more maintainable change. 

As always, I'm open to feedback :)

[Example here](https://github.com/davidsmejia/refinebio-frontend/tree/davidsmejia/support/614-horizontal-scroll-fade-option)

arielsvn:
@dvenprasad (our designer) might be able to provide more insight, but I think we wanted the buttons to grab the attention of the users intentionally. She did some tests and people used to miss them easily. The problem as it's now is that it animates too often.

Ideally we wanted to cap the number of times that the animations were fired, and have it at most once every 10 mins. That way if someone uses the app for a long period of time they might see the animation again. do you think something like this makes sense?

Also, thanks for sending this! cool fix 🔥

dvenprasad:
The ten minute cap doesn't address initial intent of the animation, which is to let people know that the table is horizontally scrollable. Instead, could we cap the number of times it animates on every page load? 

We could "slow down" the animation, i.e increase the animation length (2.5s).

Another point where it is a little annoying is where the user scrolls horizontally and then scrolls back. So if there's a way to disable the animation when the user scrolls back would be helpful. 



@arielsvn @davidsmejia what are your thoughts on this?

arielsvn:
> We could "slow down" the animation, i.e increase the animation length (2.5s).

I tried increasing the time but seems too slow IMO, it's not evident in this gif.



> Another point where it is a little annoying is where the user scrolls horizontally and then scrolls back. So if there's a way to disable the animation when the user scrolls back would be helpful.

This should be fixed on this PR, since the animation will only occur once every X mins. There's a demo of this PR deployed [here](https://deploy-preview-668--refinebio.netlify.com/). Do you think this works @dvenprasad ?

src/components/HorizontalScroll.js
```diff
@@ -3,6 +3,22 @@ import './HorizontalScroll.scss';
...
 import useVisibility from 'react-use-visibility';
 
+/**
+ * Keep track of a global timer for nudging the user to view the
+ * right scroll button. This resets every 10min starting after the
+ * first time it is animated.
+ */
+let showAnimation = true;
+let animationTimer = false;
+const didAnimate = () => {
```
arielsvn: Interesting solution! @davidsmejia is there any disadvantage of having this global variable here? I can't think of any now, but seems strange.

davidsmejia: Ya know at the time I pushed that it seemed fine. Since I was solving for keeping the value consistent across multiple pages and ```horizontalScroll``` instances. 

You'd think I never heard of ```localstorage```  before.

How about something like this...

This will fire during a leading time of 5s and then expire and re-fire after 10min. 

Thoughts?

```javascript

function ButtonRight({ onClick, disabled }) {
  const ref = useRef();
  const isVisible = useVisibility(ref.current);
  const now = (new Date()).getTime();
  const [animationTime, setAnimaionTime] = useLocalStorage(
    `horizontalScroll/animationTime`
  );
  const wasUndefined = typeof animationTime === 'undefined';
  const lastAnimationAt = animationTime || now;
  const leading = now - lastAnimationAt <= (5 * 1000);
  const trailing = now - lastAnimationAt > (10 * 60 * 1000);
  const showAnimation = leading || trailing;
  const animate = showAnimation && (isVisible && !disabled);

  React.useEffect(() => {
    if (animate && leading && wasUndefined) {
      setAnimaionTime(now);
    }
    if (animate && trailing) {
      setAnimaionTime(now);
    }
  }, [animate, wasUndefined, leading, trailing, now, setAnimaionTime]);

  return (
    <button
      type="button"
      className={classnames({
        'horizontal-scroll__right': true,
        'horizontal-scroll__disabled': disabled,
      })}
      onClick={onClick}
    >
      <div
        ref={ref}
        className={classnames({
          'horizontal-scroll__button': true,
          'horizontal-scroll__button--animate': animate,
        })}
      >
        {'>'}
      </div>
    </button>
  );
}

```

davidsmejia: Actually in theory we could also change that leading time to something that would only be true if the user saw the initial animation by biding to ```animationend``` so no risk of it firing multiple times during that (in this case 5s)

davidsmejia:
Also if you wanted you just draw attention to the scroll button without having it do a full blown harlem shake I could send over some other options:

Here is a bounce that is less attention grabby but hard to miss the first time. Also adds some cursor hints and a slight animation delay since it would give someone a change to start scanning the content before the nudge to scroll.

FYI This is branched from the old one where it fires more often just to help show the animation. I would suggest minimizing how often the nudge occurs too.

src/components/HorizontalScroll.js
arielsvn: I'm not sure I understand fully how does this works... so after the component is loaded it will wait for 5secs to trigger the animation? and then wait at least another 10mins before triggering it again?

arielsvn:
> Here is a bounce that is less attention grabby but hard to miss the first time. Also adds some cursor hints and a slight animation delay since it would give someone a chance to start scanning the content before the nudge to scroll.

Nice idea on modifying the cursor, I like that animation too. Although we tested the one we have with users so it might be risky to change it.

src/components/HorizontalScroll.js
davidsmejia: Assuming that the button is visible and not disabled...

The animation class would be applied immediately then it would be removed after 5 seconds. If the animation class was applied 10min ago or longer it would effectively reset and apply the animation class immediately (by updating the ```animationTime``` in localStorage).

arielsvn: I see, yeah I think that works! Is it possible to abstract the logic that decides whether to animate or not into a custom function/hook? Looks kind of complicated between the effect and `trailing`/`leading` variables, but that might be just me.

davidsmejia: Yeah thats a great idea. I think that would be a much more elegant solution. I'll merge in the above changes then abstract it out in a subsequent commit.

davidsmejia:
@arielsvn Could you give me some feedback on this implementation before I push. I need to test more but the Api is down at the moment.

The benefits of doing it this way would be that ```animate```could change to false and force a render in (more or less) real time. Could be useful for animating data visualizations too... down the line.

Conceptually, it's very similar to a css easing function.

```hooks.js```
```javascript
/**
 * Get a computed value based on time since passed in startTime
 * returns the value when a new evaluation occurs
* @param {*} startTime Number (time in ms) that represents when the time period starts
* @param {*} duration Number or Function that returns Number for the duration of the timePeriod(time in ms)
* @param {*} evaluator Function that takes time since period started and returns computed value
* @param {*} callback(endedAt) Optional Function that is called when the period ends

 */
export function useTimePeriod(startTime, duration, evaluator, callback) {
  const [result, setResult] = React.useState();
  const rafRef = React.useRef(false);
  const getDuration = React.useCallback((elapsed) => {
    if (typeof duration === 'number') return duration;
    if (typeof duration === 'function') return duration(elapsed);
    return Infinity; // default to forever?
  }, [duration]);

  const step = (time) => {
    const now = Date.now();
    const elapsed = now - startTime;
    if (elapsed < getDuration(elapsed)) {
      setResult(evaluator(elapsed));
      rafRef.current = requestAnimationFrame(step);
    } else {
      rafRef.current = false;
      typeof callback === 'function' && callback(now);
    }
  };

  React.useEffect(() => {
    if (!rafRef.current && startTime && Date.now() >= startTime) {
      rafRef.current = requestAnimationFrame(step);
    }
    return () => {
      rafRef.current = false
    }
  });

  return result;
}

```

and then you would use it like this

```HorizontalScroll.js```
```javascript
function ButtonRight({ onClick, disabled }) {
  const ref = useRef();
  const isVisible = useVisibility(ref.current);
  const [start, restart] = useLocalStorage('HorizontalScroll/start');
  const animate = useTimePeriod(
    start,
    10 * 60 * 1000,
    (t) => t < 5 * 1000,
    restart
  );
  // set beginning of animation time period when first seen
  React.useEffect(() => {
    if (isVisible && !disabled && !start) restart(Date.now());
  });

  return (
    <button
      type="button"
      className={classnames({
        'horizontal-scroll__right': true,
        'horizontal-scroll__disabled': disabled,
      })}
      onClick={onClick}
    >
      <div
        ref={ref}
        className={classnames({
          'horizontal-scroll__button': true,
          'horizontal-scroll__button--animate': animate,
        })}
      >
        {'>'}
      </div>
    </button>
  );
```

arielsvn:
I guess this could work, although `useTimePeriod` is a bit hard to follow could you try to explain more how does it works? Why use `requestAnimationFrame` instead of `setTimeout`?

davidsmejia:
@arielsvn The hook works by creating a loop that checks against the passed in start time until you get to the the duration. Then it calls the passed in callback. The hooks returned value (the returned value from the passed in evaluator function) is updated/called as time progresses so this allows the component to render the returned value as it changes... as opposed just whatever it is at the time of ```useTimePeriod``` being called in the component.

Also ```requestAnimationFrame``` was used for keeping the checks against the elapsed time continuous without calling it more often than would be practical. 

I see your concern that it's more complicated than it has to be... another way to solve this would be the following... but it the hook doesn't trigger any state changes as time progresses. So in this context ```animate``` would be whatever it's value was when executed until it was rendered again when ```isVisible``` or ```disabled``` changes value. 

This stops the animation from starting over when ```disabled``` gets toggled true/false.
```HorizontalScroll.scss```
```css
  &__disabled {
    visibility: hidden;
  }
```

```HorizontalScroll.js```
```javascript
function useStartTimer(start, timeout, callback) {
  if (!start) return 0;
  const time = Date.now() - start;
  if (time >= timeout) typeof callback === 'function' && callback(Date.now());
  return time;
}

function ButtonRight({ onClick, disabled }) {
  const ref = useRef();
  const isVisible = useVisibility(ref.current);
  const [start, restart] = useLocalStorage('HorizontalScroll/timeout');
  const time = useStartTimer(start, 10 * 60 * 1000, restart);
  if (!start && isVisible && !disabled) restart(Date.now());
  const animate = time < 5 * 1000;

  return (
    <button
      type="button"
      className={classnames({
        'horizontal-scroll__right': true,
        'horizontal-scroll__disabled': disabled,
      })}
      onClick={onClick}
    >
      <div
        ref={ref}
        className={classnames({
          'horizontal-scroll__button': true,
          'horizontal-scroll__button--animate': animate,
        })}
      >
        {'>'}
      </div>
    </button>
  );
}

```

davidsmejia:
And yeah in that last example you wouldn't really need to put it in a side effect because it doesn't change anything. So maybe something like this instead...

I'm still thinking about who should own the state for ```start``` and ```restart```

```javascript
function useTimerStateDefered(start, restart, duration, canStart) {
  const time = Date.now() - start;

  React.useEffect(() => {
    const now = Date.now();
    if (!start && canStart) restart(now)
    if (time >= duration) restart(now);
  }, [canStart, duration, restart, start, time]);

  return (!start) ? 0 : time;
}

function ButtonRight({ onClick, disabled }) {
  const ref = useRef();
  const isVisible = useVisibility(ref.current);
  const [start, restart] = useLocalStorage('HorizontalScroll/animate');
  const time = useTimerStateDefered(
    start,
    restart,
    10 * 1000,
    () => isVisible && !disabled
  );

  const animate = time < 1.1 * 1000;

render...

```

arielsvn:
I think we might have gone too deep into the rabbit hole. I had a discussion with @dvenprasad that might bring some light into this and simplify things. Specifically, we don't need to cap the number of animations every 10 mins. I think I was the one that suggested that in the beginning 😬 

The major concern with the shaking is that it happens multiple times when scrolling horizontally _(image from Deepa's comment above)_



As long as we prevent this and add a 1 sec delay before the animation starts we should be fine. 

What do you think @davidsmejia? This is kind of bittersweet, since we spent some time discussing a solution for this that was in a different direction.

davidsmejia:
@arielsvn yeah sounds good, I'm going to go ahead and close this PR and create a new one that addresses those issues.

Issue #613, opened 2019-04-23
Add "loading" state for buttons


dvenprasad:
Sometimes adding samples to dataset takes a while, especially when the number of samples is large. Currently "loading" is indicated by the progress bar on the top but it does not make it clear to users that refine.bio is working on adding those samples to their datasets. 

Adding a loading state to the button makes it clear and makes the system feel more "responsive" to users.

@arielsvn  Reference for loading state: https://coreui.io/docs/components/loading-buttons/

wvauclain:
Closed by #688

PR #670, opened 2019-06-22
adds loading state for onClicks that return a promise

davidsmejia:
## Issue Number

#613 

Sometimes on slow connections async calls take more than a beat to complete. This PR adds a wrapper method to a ```Button``` component's ```onClick``` property that checks if the passed in function returns a promise. If it doesn't or it's not a function it's business as usual. But if it does return a promise the button will display a little spinner and hide the contents of the button. It uses ```visibility:hidden``` and ```color: transparent``` in the button to keep it's layout the same during the loading state.

Additionally,
I made the assumption that multiple clicks should be handled after the previous click has been returned so it's essentially disabled during the load state. This should be confirmed as there might be a place in the app now or in the future where this assumption is wrong.

Also this is currently opt-in by default but could be made opt-out by default too.

One last thing. Maybe on error there should be a shake event to communicate that something went wrong? 

## Types of changes

## Functional tests

I click a bunch of buttons that return promises(like adding or removing something from your dataset) and ones that don't (like showing a modal or the scroll back to top button).

## Checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -136,7 +136,8 @@ $default-active-shadow: 0 3px 4px 0 rgba($black, 0.3);
       color: transparent;
     }
 
-    .button--indicator {
+    &:before {
+      content: '';
       visibility: visible;
       display: inline-block;
       vertical-align: baseline;
@@ -164,7 +165,7 @@ $default-active-shadow: 0 3px 4px 0 rgba($black, 0.3);
 
     &.button--plain,
     &.button--link {
-      .button--indicator {
+      &:before {
         top: auto;
         right: 0;
         animation-name: border-color-change, loading-spin-inline;
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -126,7 +126,86 @@ $default-active-shadow: 0 3px 4px 0 rgba($black, 0.3);
     font-size: 16px;
     padding: 8px 32px;
   }
+
+  &--loading {
+    position: relative;
+    color: transparent;
+    &:hover,
+    &:active,
+    &:focus {
+      color: transparent;
+    }
+
+    .button--indicator {
+      visibility: visible;
+      display: inline-block;
+      vertical-align: baseline;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      margin: 0;
+      width: 10px;
+      height: 10px;
+      background-color: transparent;
+      border-top: 1px solid transparent;
+      border-right: 1px solid transparent;
+      border-left: 1px solid transparent;
+      border-bottom: 1px solid transparent;
+      border-radius: 50%;
+
+      animation: border-color-change, loading-spin-center;
+      animation-duration: 1.2s;
+      animation-iteration-count: infinite;
+
+      + * {
+        visibility: hidden;
+      }
+    }
+
+    &.button--plain,
+    &.button--link {
+      .button--indicator {
+        top: auto;
+        right: 0;
+        animation-name: border-color-change, loading-spin-inline;
+      }
+    }
+  }
 }
 
 .button__icon {
+}
+
+@keyframes border-color-change {
+  from,
+  to {
+    border-top-color: white;
+    border-right-color: $azure;
+    border-bottom-color: white;
+    border-left-color: $azure;
+  }
+  50% {
+    border-top-color: $azure;
+    border-right-color: white;
+    border-bottom-color: $azure;
+    border-left-color: white;
+  }
+}
+
+@keyframes loading-spin-center {
+  from {
+    transform: translate(-50%, -50%) rotate(0deg);
+  }
+  to {
+    transform: translate(-50%, -50%) rotate(360deg);
+  }
+}
+
+@keyframes loading-spin-inline {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
 }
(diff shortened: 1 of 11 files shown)


src/components/Button/Button.scss
```diff
@@ -126,7 +126,86 @@ $default-active-shadow: 0 3px 4px 0 rgba($black, 0.3);
...
+  &--loading {
+    position: relative;
+    color: transparent;
+    &:hover,
+    &:active,
+    &:focus {
+      color: transparent;
+    }
+
+    .button--indicator {
```
arielsvn: can we achieve the same with a pseudo-element? `:before`? To avoid having to add a new element

src/components/Button/index.js
```diff
@@ -6,20 +6,49 @@ const Button = ({
...
+  className = '',
   children,
   type = 'button',
+  onClick,
   ...props
 }) => {
+  const loadingRef = React.useRef(false);
+  const promiseRef = React.useRef();
+
+  const buttonClick = () => {
```
arielsvn: can this be simplified with `async`?

```js
const [loading, setLoading] = React.useState(false);
const buttonClick = async () => {
  setLoading(true);
  await onClick();
  setLoading(false)
}
```

arielsvn:
Also, initially we were thinking on an [animation like this one](https://coreui.io/docs/components/loading-buttons/):



But I like the one you added, specially because it doesn't modify the size of the buttons. What do you think @dvenprasad, does this one works? [demo here](https://deploy-preview-670--refinebio.netlify.com/)

src/components/Button/Button.scss
davidsmejia: Yeah I had that in there before I pushed but I wasn't sure if any other subclasses of button would have their own rules for ```:before``` or ```:after```. If you think thats safe I can definitely remove the element and change the rule.

arielsvn: Yes, it should be. All button styles are on Buttons.scss

davidsmejia: Sounds good 👍

dvenprasad:
Yep. This is so much better. 🏆



On Mon, Jun 24, 2019 at 4:55 PM -0500, "Ariel Rodriguez Romero" <notifications@github.com<mailto:notifications@github.com>> wrote:


Also, initially we were thinking on an animation like this one<https://coreui.io/docs/components/loading-buttons/>:

[2019-06-24 16 48 26]<https://user-images.githubusercontent.com/1882507/60054540-7e55d980-96a0-11e9-8aa8-4a84689eb5c4.gif>

But I like the one you added, specially because it doesn't modify the size of the buttons. What do you think @dvenprasad<https://github.com/dvenprasad>, does this one works? demo here<https://deploy-preview-670--refinebio.netlify.com/>

src/components/Button/index.js
davidsmejia: Keeping track of the promise just helps prevent multiple clicks from triggering async actions until the previous one has resolved and I think we would have to move ```setLoading``` into a ```React.useEffect```

If there is a ```Promise.all``` polyfill we could do this

```javascript
  const loadingRef = React.useRef(false);
  const promiseRef = React.useRef();
  const updateRefs = (isLoading = false, promise = null) => {
    loadingRef.current = isLoading;
    promiseRef.current = promise;
  }

  const buttonClick = () => {
    if (typeof onClick !== 'function') return onClick;
    if (loadingRef.current) return promiseRef.curent;

    const result = onClick();
    updateRefs(true, result);
    Promise.all([result]).then(() => updateRefs());
    return result;
  }
```

If not...
Do you think this is clearer?

```javascript
  const loadingRef = React.useRef(false);
  const promiseRef = React.useRef();
  const updateRefs = (isLoading = false, promise = null) => {
    loadingRef.current = isLoading;
    promiseRef.current = promise;
  }

  const buttonClick = () => {
    if (typeof onClick !== 'function') return onClick;
    if (loadingRef.current) return promiseRef.curent;

    const result = onClick();
    const isPromise = result && typeof result.then === 'function';

    if (isPromise) {
      updateRefs(true, result);
      result.then(() => updateRefs());
    }
    return result;
  }
```

arielsvn: hmm but what if we only execute `onClick` when `loading===false`? That way we ensure that async actions are only triggered once at a time.

```
const [loading, setLoading] = React.useState(false);
const buttonClick = async () => {
  if (loading) return;
  setLoading(true);
  await onClick();
  setLoading(false)
}
```

I assumed no clicks would be executed since the button gets disabled while a promise is being resolved. Does this works?

davidsmejia: Yeah I think this would work. 

Some implications though worth testing/considering...
- If the ```onClick``` event did something that took a long time to compute would it display a loading state right? 
That does seem to have advantages over just checking for a promise. For instance, if you were doing something computationally expensive on the front end?
- We might need to put it in a ```useEffect``` incase it set state after the button was unmounted? Or maybe just keep it as a ```useRef```

arielsvn: > If the onClick event did something that took a long time to compute would it display a loading state right?

Yes, `loading` should be `true` while the promise is being executed.

> We might need to put it in a useEffect incase it set state after the button was unmounted?

Yes, that seems like a good solution. Good catch!

davidsmejia:
Going to close this and open a new PR from a clean branch from dev. I absentmindedly rebased and pushed so I believe this will rewrite history.

PR #671, opened 2019-06-25
Queued Actions Middleware

davidsmejia:
## Issue Number

Waiting on adressing #613

This adds middleware ```src/store/queueMiddleware.js``` for queueing requests, specifically to benefit the experience on slower connections that have to wait longer for the asynchronous updates to the dataset which overrides previous (and still in the process of being executed) requests - such as adding a couple samples to your dataset. (Reference the attached screenshots.)

Apart from adding the middleware itself, it applies changes to ```state/download/actions.js``` to implement the new middleware. Such as: 

```javascript

/**
 * Takes an array of experiment objects and adds to users dataset via endpoint
 * @param {object} dataSetSlice
 */
export const addSamples = dataSetSlice => async dispatch =>
  dispatch({
    queueChannel: 'dataSetUpdateOperation',
    queuedAction: dataSetUpdateOperation(dataSet =>
      new DataSetManager(dataSet).add(dataSetSlice)
    ),
  });
```

## Types of changes

## Functional tests

This helps to test if you open developer tools in chrome -> options -> more tools -> network conditions -> and test with either ```Fast 3G``` or ```Slow 3G```. It still occurs on fast connections if you click around quickly but throttling your connect makes it considerably easier to replicate. I mean it's a race condition so it's naturally unpredictable. 

## Checklist

## Screenshots
Original:

With Queues:

### DIFF
--- a/src/store/queueMiddleware.js
+++ b/src/store/queueMiddleware.js
@@ -1,0 +1,49 @@
+/**
+ * This middleware allows for dispatched actions to be called successively, FIFO.
+ * Use by passing and action of type {queuedAction: func, queueChannel: 'CHANNEL_NAME_ACTION_TYPE'})
+ * The action passe will be defered until
+ */
+
+const queueMiddleware = ({ dispatch, getState }) => {
+  const queues = {};
+  return next => async action => {
+    const { queuedAction, queueChannel } = action;
+    const isFunction = typeof queuedAction === 'function';
+    const isObject = typeof queuedAction === 'object';
+    const isQueueChannel = typeof queueChannel === 'string';
+    const didQueue = (isFunction || isObject) && isQueueChannel;
+
+    if (!didQueue) return next(action);
+
+    const queue = Array.isArray(queues[queueChannel])
+      ? queues[queueChannel]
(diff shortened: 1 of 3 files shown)


src/store/queueMiddleware.js
```diff
@@ -0,0 +1,49 @@
...
+
+    const queue = Array.isArray(queues[queueChannel])
+      ? queues[queueChannel]
+      : (queues[queueChannel] = []);
+
+    const promise = new Promise(async (resolve, reject) => {
+      const resolveQueuedAction = async () => {
+        if (typeof queuedAction === 'function') {
+          try {
+            const result = await queuedAction(dispatch, getState);
```
arielsvn: is it the same to do `await dispatch(queuedAction)` ?

arielsvn:
Tested locally, I think we might still be running into some race conditions



Loved how the buttons show as pending until they resolve :) maybe this change is related to the race condition, I remember we changed `useLoader` for that.

davidsmejia:
Going to close this and re-implement from the more recent version of dev

PR #688, opened 2019-07-10
Add Loading State For Buttons

davidsmejia:
## Issue Number

#613 

Same styles as #670 but logic is extracted into a hook for portability.

## Types of changes

## Functional tests

I tested this on every implementation of button I could find.

## Checklist

### DIFF
--- a/src/common/hooks.js
+++ b/src/common/hooks.js
@@ -100,12 +100,12 @@ export function useWaitForAsync(func) {
   const mountedRef = React.useRef(true);
   const [waiting, setWaiting] = React.useState(false);
 
-  React.useEffect(() => {
-    mountedRef.current = true;
-    return () => {
+  React.useEffect(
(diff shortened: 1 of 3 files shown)


src/common/hooks.js
```diff
@@ -90,3 +90,29 @@ export function useLocalStorage(key, initialValue) {
...
+export function useWaitForAsync(func) {
+  const mountedRef = React.useRef(true);
+  const [waiting, setWaiting] = React.useState(false);
+
+  React.useEffect(() => {
+    mountedRef.current = true;
+    return () => {
+      mountedRef.current = false;
+    };
+  }, [waiting]);
```
arielsvn: is this dependency needed here?

davidsmejia: Oh good catch, I will clean this up

PR #689, opened 2019-07-11
Arielsvn/639 downloadable samples filter

### LINKED ISSUE #639: Add option to filter "no downloadable samples" out
I was searching for a small experiment to use for testing so I ordered by "least No. of samples" and all I see is pages of experiments with no downloadable samples. I think we should provide an option to filter all results like that out. I also think it makes sense outside of this context because unless you're looking for specific experiments and want to see that they're there but don't have downloadable samples, you probably wanna filter all the experiments you can't add to your dataset anyway.

arielsvn:
## Issue Number

close #639 

Adds a new filter to `show experiments with 0 downloadable samples`, and by default only search for experiments with downloadable samples.

## Types of changes

## Functional tests

Tested locally.

## Checklist

## Screenshots

image


src/pages/Search/SearchContext.js
```diff
@@ -0,0 +1 @@
+export function SearchContext({}) {}
```
wvauclain: Is this being used anywhere?

arielsvn:
I pushed a few more changes @wvauclain in case you want review again.

Also here's a [demo](https://deploy-preview-689--refinebio.netlify.com/) @dvenprasad 

image

dvenprasad:
@arielsvn Can we increase the padding/margin between `Showing etc..` and `Hide experiments` to 16px. It looks like it's really jammed in there.

Issue #691, opened 2019-07-12
Current dataset is removed after refreshing


arielsvn:
![2019-07-12 11 15 33](https://user-images.githubusercontent.com/1882507/61138971-6a7ae780-a496-11e9-95e8-a7e9c25c2953.gif)

wvauclain:
It looks like this happens because `dataset-id` is somehow removed from Local Storage when you click on the `My Dataset` button.

wvauclain:
Though this also seems to happen when you refresh the page, which causes a related bug. You can refresh once on the search page without losing your dataset, but it gets lost when you refresh for the second time.

davidsmejia:
If you add 
```javascript
console.error('id', action.data.id, 'datasetid', action.data.dataSetId);
``` 
to ```src/state/download/reducer.js```
You can see that the two values flip flop depending on the type of state change that occurs.
Which ultimately makes the persist middleware remove the key which makes you lose your dataset session as @wvauclain pointed out.

wvauclain:
Fixed by #694

PR #692, opened 2019-07-12
dataSetId is id

davidsmejia:
## Issue Number

#691 

Adds ```|| data.id``` incase the passed in object's id for dataset isn't ```dataSetId``` in the download actions reducer.

## Types of changes

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist

### DIFF
--- a/src/state/download/actions.js
+++ b/src/state/download/actions.js
@@ -26,7 +26,7 @@ export const updateDownloadDataSet = data => ({
   type: 'DOWNLOAD_DATASET_UPDATE',
   data,
   persist: {
-    dataSetId: data.dataSetId,
+    dataSetId: data.dataSetId || data.id,
   },
 });
 


src/state/download/actions.js
```diff
@@ -26,7 +26,7 @@ export const updateDownloadDataSet = data => ({
   type: 'DOWNLOAD_DATASET_UPDATE',
   data,
   persist: {
-    dataSetId: data.dataSetId,
+    dataSetId: data.dataSetId || data.id,
```
arielsvn: Thanks for sending this, it fixes the issue! 🔥 

Is it possible to find where this is called and pass the `dataSetId` there? It would be nice if we could have a single parameter instead of using both.

davidsmejia: Yeah I can look into it more, I have a hunch that its coming back from the API inconsistently. So I'm not sure at what level you want to sanitize the id. I left another comment on the issue thread that helps reveal when the value changes. 

Do you want to merge this or wait and address the inconsistency before it gets to the downloads reducer or maybe at the reducer we can check if dataSetId is empty and try id... thoughts?

arielsvn: hmm I think the problem might be related to how we parse the api response, since the api always returns `id` and we use `dataSetId` in the code.

Yeah let's get this merged and we can find a better solution later.

Issue #731, opened 2019-08-21
Sitemap.xml is not being generated


davidsmejia:
After changing the API urls to include a version the sitemap.xml is no longer being generated on deploys.

Next step would be to add v1 to the ```cacheBackend.js``` script or possibly add ```latest``` to the api so endpoints like these will always just go to the latest version.

davidsmejia:
As it currently is being generated. For the 50k experiments, it makes 100 api requests for ~500 experiments per request. It also looks like almost all of them are erring out. I think the best move would be to create a separate api endpoint that could just return the accession_code, title, and maybe urls in ```created_at``` order ascending so we could possibly just get from the last one and insert into the sitemap by default instead of asking for all 50k + going forward. 

This would also allow for us to add autocomplete / suggestions into search.

davidsmejia:
This might also be a more performant option for the future. Just putting it here for future reference.
 
https://github.com/taskrabbit/elasticsearch-dump

arielsvn:
But with another API endpoint wouldn't we be in the same situation? Why are the requests failing?

The number of accessions could be way bigger if we also included the samples in the sitemap.

davidsmejia:
@arielsvn I think the problem is that Django can't handle the 1,000 requests in such a short amount of time. So my thinking now is that it would be best to query ES directly as save the results to s3. I'm not sure where that command should be executed from but it would have to have access to the ES endpoint.

PR #735, opened 2019-08-25
Generate Sitemaps on Deploy: The Saga of SEO

davidsmejia:
## Issue Number

#731 

this will query the /search endpoint sequentially until there are no more results, ordering experiments by ```id``` rather than ```created_at``` (since it's not currently exposed as an ordering option) 
it takes a little over 1 min to run top to bottom

***I also addressed the versioning of the cachebackend script, which I think was silently broken***

- updates the sitemap npm package
- pulls out the sitemap generation from the cache backend 
- creates a new script for generating a sitemap index file and various sitemaps from a list of urls
- modifies circle ci script to generate staging or production sitemaps
- change robots.txt to use new sitemap index

Note: staging sitemaps should be generated and manually verifiable but not indexed

Further thoughts for future improvements:
- add better SEO in the ```<head>``` of experiments page (should be addressed before deploying this)
- check against s3 and pass in existing sitemaps so only those with <50k are regenerated (revisit sometime before 100k experiments)
- add ```changeFrequency``` to never when processed samples is equal to total sample
- add ```changeFrequency``` to weekly/monthly when there are unprocessed samples
- add more resources! samples can link to their experiments page or a samples detail page?
- possibly consider moving this to a lambda function so its regenerated as every X experiments are surveyed

## Types of changes

## Functional tests

ran from yarn with no params which works fine, it pulls from the prod API and creates 1 sitemap-```index.xml``` and two ```sitemap-#.xml``` files with a total of 50k+ urls for experiments and saves it to ```/public```

## Checklist

### DIFF
--- a/.circleci/deploy.sh
+++ b/.circleci/deploy.sh
@@ -21,8 +21,10 @@ branch=$(get_master_or_dev)
 
 if [[ $branch == "master" ]]; then
     base_host="refine.bio"
+    host_name="https://www.refine.bio"
 elif [[ $branch == "dev" ]]; then
     base_host="staging.refine.bio"
+    host_name="https://staging.refine.bio"
 else
     echo "Why in the world was update_docker_img.sh called from a branch other than dev or master?!?!?"
     exit 1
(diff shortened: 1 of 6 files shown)


davidsmejia:
Probably should add this for production deploys. Curling these should request that the sitemap has been updated and can be reindexed automagically.

- ```http://www.google.com/ping?sitemap=https://www.refine.bio/sitemap-index.xml```
- ```http://www.bing.com/webmaster/ping.aspx?siteMap=https://www.refine.bio/sitemap-index.xml```

Issue #721, opened 2019-08-14
Switch over forms on refine.bio to Hubspot


dvenprasad:
There are 3 places where forms are embedded on refine.bio collection email addresses. 

1. refine.bio landing page

2. https://www.refine.bio/species-compendia

3. https://www.refine.bio/download?start=true ( When user downloads the dataset) 
![Screen Shot 2019-08-14 at 3 11 35 PM](https://user-images.githubusercontent.com/15315514/63049059-d55c7b80-bea5-11e9-80f4-c40764b41d02.png)
The email addresses collected from this need to either be "tagged" as refine.bio downloaders or stored as a separate list so I can automate follow-ups.

dvenprasad:
Note: The forms on hubspot will probably need to be modified to match our forms.

dvenprasad:
This is coming together nicely. 
Couple of notes:

1. Let's remove the input field once they subscribe. 
2. Copy: `Thank you for subscribing. Updates will be sent to <email@email.com>`

PR #746, opened 2019-08-30
Hide email subscription form after submit

davidsmejia:
## Issue Number

#721 

Hides the email form after submit 
Adds new success message

## Types of changes

## Functional tests

tested locally 

## Checklist

## Screenshots

Screen Shot 2019-08-30 at 12 03 21 PM

### DIFF
--- a/src/pages/Main/EmailSection.js
+++ b/src/pages/Main/EmailSection.js
@@ -1,18 +1,21 @@
 import React from 'react';
+import classnames from 'classnames';
+import * as Yup from 'yup';
+import { Formik, Field, Form } from 'formik';
 import { Ajax } from '../../common/helpers';
+import Button from '../../components/Button';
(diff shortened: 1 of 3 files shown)


src/pages/Main/EmailSection.js
```diff
@@ -24,8 +24,8 @@ export default function EmailSection() {
   const handleSubmit = async e => {
     e.preventDefault();
     try {
-      const result = await Ajax.post(formUrl, getFormData());
-      setSubscribed(result.inlineMessage);
+      await Ajax.post(formUrl, getFormData());
+      setSubscribed(true);
     } catch (err) {
       // Show error message setError(true); "Please submit a valid e-mail address."
```
arielsvn: Can we show a validation error if the email is not valid? I think we use Formik in other forms to handle this.

Issue #748, opened 2019-09-03
disallowed urls should also have metatag on page


davidsmejia:
In robots.txt there is the following:
```
Disallow: /download
Disallow: /dataset/*
```

We should add:
```
<meta name="robots" content="noindex, nofollow">
```

PR #750, opened 2019-09-03
adds meta tags noindex, nofollow

davidsmejia:
## Issue Number

#748 

Adds meta tags to prevent indexing of downloads/dataset pages

## Types of changes

## Functional tests

Loaded the pages

## Checklist

### DIFF
--- a/src/pages/Downloads/DataSet/index.js
+++ b/src/pages/Downloads/DataSet/index.js
@@ -56,6 +56,7 @@ export default function DataSet({
     <div>
       <Helmet>
         <title>Dataset - refine.bio</title>
+        <meta name="robots" content="noindex, nofollow" />
         <meta
           name="description"
           content="Explore and download this custom harmonized childhood cancer transcriptome dataset."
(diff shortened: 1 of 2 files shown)


arielsvn:
What's the difference between adding these tags in the headers vs adding them in `robots.txt`?

davidsmejia:
If a page is ever linked, google might try to index it. Even if it disagrees with our robots.txt google says it would like to verify. Its more of a warning than an error. 


> Warning
> Pages with a warning status might require your attention, and may or may not have been indexed, according to the specific result.
> 
> Indexed, though blocked by robots.txt: The page was indexed, despite being blocked by robots.txt (Google always respects robots.txt, but this doesn't help if someone else links to it). This is marked as a warning because we're not sure if you intended to block the page from search results. If you do want to block this page, robots.txt is not the correct mechanism to avoid being indexed. To avoid being indexed you should either use 'noindex' or prohibit anonymous access to the page using auth. You can use the robots.txt tester to determine which rule is blocking this page. Because of the robots.txt, any snippet shown for the page will probably be sub-optimal. If you do not want to block this page, update your robots.txt file to unblock your page.

PR #769, opened 2019-10-07
Enable automatic deploys on master and dev

### LINKED ISSUE #754: Deploy automatically when master or dev are updated
We always deploy prod and staging when pushing a tag, but I think we did that to match the backend that had a more complicated process.

Since we have both `master` and `dev` protected we could make the deploys automatic after a PR is merged on those branches.

arielsvn:
## Issue Number

close #754 

Enable automatic deploys on `master` and `dev`.


arielsvn:
AUtomatic deploys are working!

image

Issue #758, opened 2019-09-16
Compendium Page -> Normalized compendium / quantpendium


cgreene:
Based on https://github.com/AlexsLemonade/refinebio/issues/1587 it's clear that this is desired. We should update the compendium page to continue to provide the headline normalized compendium, but also to provide - either below the fold or after changing selection, a collection of all of the `quant.sf` files that we've generated organized by species.

dvenprasad:
Species Compendia and Quant.sf Compendia will be tucked under the Compendium Page. 

Copy to contextualize the compendium:
`refine.bio Compendium packages the data processed by refine.bio pipelines for flexible and broad use by computational biologists and data scientists`

Things that need to go on the Quant.sf Compendia page/section
- Copy: What is it? How was it processed?
- Download mechanism
- Customize the download via API
- Explicitly mention that files they receive are NOT normalized or aggregated.  

Also, do we have information/links to close the loop when users download quant compendia like we do for species compendia? or can we skip it for now? @cgreene @jaclyn-taroni 
![Screen Shot 2019-09-16 at 4 37 00 PM](https://user-images.githubusercontent.com/15315514/64991639-3a89ff00-d8a0-11e9-91b9-a0ab735a2621.png)



## Low Fi:
![quant-compendia-lowfi-annotated](https://user-images.githubusercontent.com/15315514/64991358-ba639980-d89f-11e9-85a2-4e9266c3d5a9.jpg)

dvenprasad:
Mockups:
![compendium-species-compendia-no-examples](https://user-images.githubusercontent.com/15315514/65347527-36febc80-dbad-11e9-8026-b8cbc737196b.png)

![compendium-rna-seq-compendia](https://user-images.githubusercontent.com/15315514/65347528-36febc80-dbad-11e9-8a79-1c10982e71c1.png)



Note: The copy under `Uniformly Processed` needs to be replaced with appropriate copy once the docs are ready. 

Invision Links:
https://invis.io/9SNAJ70ME4D#/384794314_compendium-Species-Compendia-No-Examples

https://invis.io/9SNAJ70ME4D#/384608890_compendium-Rna-Seq-Compendia

Assets:
pipeline illustration
https://invis.io/a/E58AFMU0R6FZ

RNA-seq sample compendia icon
https://invis.io/a/VE8AF3FY3982

I've left comments on the mockup with links for the hyperlinks and about transition.
@davidsmejia let me know if you need anything else from me.

[...]

PR #773, opened 2019-10-09
compendium and RNA-seq Sample compendium

davidsmejia:
## Issue Number

#758 

- Implements the new compendium and RNA-seq Sample Compendia 
- ThemeContext.js
- DownloadCompendia takes a filter option which allows for arbitrary filtering for all of your future compendia needs
- Header now has a DropDown links component which takes an array of objects with keys ```title``` to display for the link and ```location``` which is a ```react-router-dom``` [object or string](https://reacttraining.com/react-router/web/api/Link)

ThemeContext.js:
In order to solve the problem with the various headers on the site overriding the new dropdown navigation elements I have implemented a theme context that allows pages to tell other components which color they would like them to be... or handle a specific color request. It has a simple interface which uses React Hooks under the hood.
ie ```import {useTheme} from './common/ThemeContext.js';```
and then if you want to set a theme:
```useTheme('light')```
and if you want to handle themes:
```const [theme] = useTheme()```
then ``` ... if (theme.thePropertyIcareAbout === 'light') makeThingLight()```
so any component can decide how if and how it should handle that... 

Just a thought:
Also.. ThemeContext.js should probably go to ```src/common/hooks.js``` and use a config file from ```src/common/themes.js``` instead

## Types of changes

## Functional tests

Yarn tests, can download 

Note: large files seem to take a long time and there is an issue with Xorigin cookies that is probably coming from AWS

## Checklist

## Screenshots

Screen Shot 2019-10-09 at 7 07 13 PM

### CODE REVIEW

File: src/pages/Compendia/index.js [hunk 1]
```diff
@@ -0,0 +1,76 @@
...
+import classnames from 'classnames';
+import { Link, withRouter, Redirect } from 'react-router-dom';
+import './Compendia.scss';
+import PointsBackground from './PointsBackground';
+import NormalizedCompendia from './NormalizedCompendia';
+import RNASeqSampleCompendia from './RNASeqSampleCompendia';
+import EmailSection from '../Main/EmailSection';
+import { useTheme } from '../../common/ThemeContext';
+
+const Compendia = function Compendia({ location, replace }, ...props) {
```
arielsvn: What is `...props` here for? Also, you could probably remove the variable declaration

src/pages/Main/Main.scss
```diff
@@ -90,14 +86,12 @@
...
+      box-shadow: 0 3px 20px 0 rgba(0, 0, 0, 0.2);
       overflow: hidden;
 
-      @include tablet {
-        padding: 10 * 8px 0 0;
-      }
-
       @include desktop() {
         overflow: visible;
+        min-height: 487px;
```
arielsvn: Can we use a multiple of `8px` here? `480` maybe?

src/components/Header/index.js
```diff
@@ -13,23 +13,17 @@ import SideMenu from '../SideMenu';
...
 import githubCorner from './github-corner.svg';
+import { push } from '../../state/routerActions';
 
-/**
- * In general this is a bad approach, where the header component knows about other pages.
- * But in the future we'll unify the header styles and have a single color. So it doesn't make
- * much sense to invest a lot of time improving this.
- */
-function shouldInvertColors(pathname) {
-  return ['/', '/about', '/species-compendia'].includes(pathname);
```
arielsvn: 🔥 🔥 loving the `useTheme` approach for removing this atrocity!

src/common/ThemeContext.js
```diff
@@ -0,0 +1,42 @@
...
+    if (themeName) setTheme(themeName);
+    return () => {
+      if (themeName) setTheme('default');
+    };
+  }, [themeName, setTheme]);
+
+  return [theme, setTheme];
+};
+
+export { themes, ThemeContext, ThemeProvider, useTheme };
```
arielsvn: What do you think of making `useTheme` private here and exporting two more specific hooks: `useThemeInverted()` and `useThemeLight()`?

src/common/ThemeContext.js
```diff
@@ -0,0 +1,42 @@
...
+const themes = {
+  default: {
+    header: 'default',
+  },
+  inverted: {
+    header: 'inverted',
+  },
+  light: {
+    header: 'light',
+  },
```
arielsvn: Why have the specific value for `header` here? Should the `Header` component know how to display itself depending just on the active theme?

src/components/Header/index.js
```diff
@@ -125,6 +136,80 @@ const HeaderLink = ({ to, onClick, children, location, activePath = [] }) => {
   );
 };
 
+let HeaderDropDownLink = ({
+  to = [],
+  onClick,
+  children,
+  location,
+  activePath = [],
```
arielsvn: `activePath` is needed?

src/pages/Compendia/DownloadCompendia.js
```diff
@@ -0,0 +1,185 @@
...
+  state => ({
+    token: state.token,
+  }),
+  { createToken, push }
+)(DownloadCompendia);
+export default DownloadCompendia;
+
+const filterForLatestCompendia = (data, organism = false) => {
+  data.forEach(c => {
+    c.organism_name = c.result.annotations[0].data.organism_name;
```
arielsvn: Is there a way to achieve the same here without [mutating the parameter](https://github.com/airbnb/javascript#functions--mutate-params)?

src/components/Header/index.js
```diff
@@ -125,6 +136,80 @@ const HeaderLink = ({ to, onClick, children, location, activePath = [] }) => {
...
+          </ul>
+        </li>
+      </ul>
+    </li>
+  );
+};
+
+HeaderDropDownLink = connect(
+  null,
+  { push }
```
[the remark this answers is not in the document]
davidsmejia: i think we can remove this 👍

src/pages/Compendia/DownloadCompendia.js
```diff
@@ -0,0 +1,185 @@
...
+    c => c.organism_name === organism.organism_name
+  );
+  if (downloadableFile && downloadableFile.download_url) {
+    return downloadableFile.download_url;
+  }
+  // throw error? or alert slack
+  return '';
+}
+
+const getPrettyFileSize = size => {
```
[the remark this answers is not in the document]
davidsmejia: removed

src/pages/Compendia/index.js
davidsmejia: removed

src/components/Header/index.js
davidsmejia: yeah i just forgot to implement it :) good catch!

src/pages/Compendia/DownloadCompendia.js
```diff
@@ -0,0 +1,185 @@
...
+
+  const organismNames = uniq(data.map(c => c.organism_name));
+  const filtered = [];
+
+  for (const organismName of organismNames) {
+    const organismCompendium = data
+      .filter(c => c.organism_name === organismName)
+      .sort((x, y) => y.compendia_version - x.compendia_version);
+    filtered.push(organismCompendium[0]);
+  }
```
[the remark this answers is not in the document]
davidsmejia: will do

src/pages/Main/Main.scss
davidsmejia: yup

cgreene:
Design discussed in #776

src/common/ThemeContext.js
davidsmejia: The idea here is that each theme can be composed of different colors for each component.

src/common/ThemeContext.js
davidsmejia: I've updated this to use ```useTheme(themes.light)``` you can ```import {themes, useTheme } from 'ThemeContext' ```

src/api/compendia.js
```diff
@@ -0,0 +1,33 @@
+import { Ajax } from '../common/helpers';
+
+export async function fetchCompendiaData(additionalFilters = {}, token) {
+  try {
+    const filters = {
+      ...{
```
arielsvn: You can write these properties directly instead of spreading them?

src/api/compendia.js
```diff
@@ -0,0 +1,33 @@
...
+        ordering: 'primary_organism__name',
+        limit: 1000,
+      },
+      ...additionalFilters,
+    };
+
+    const values = Object.entries(filters).map(([key, value]) => {
+      return `${key}=${value}`;
+    });
+    const endpoint = `/v1/compendia/?${values.join('&')}`;
```
arielsvn: You can also pass `filters` as the second parameter of `Ajax.get`, and it'll do something like this.

src/api/compendia.js
```diff
@@ -0,0 +1,33 @@
...
+
+    const response = await Ajax.get(endpoint, null, tokenObject);
+
+    return response.results;
+  } catch (e) {
+    return [];
+  }
+}
+
+export async function fetchCompendium(token, selected) {
```
arielsvn: what if we pass the compendium `id` as a parameter, instead of `selected`?

.eslintrc.js
```diff
@@ -62,6 +62,9 @@ module.exports = {
     'prefer-destructuring': 'warn',
     'no-param-reassign': 'warn',
 
+    // this should be fixed in header/headerA
+    'jsx-a11y/no-noninteractive-element-interactions': 'warn',
```
arielsvn: I think we talked about this, but I can't remember well... Did this caused issues?

src/pages/Compendia/DownloadCompendia.js
```diff
@@ -0,0 +1,112 @@
...
+  title,
+  filter,
+  children,
+}) => {
+  const agreedToTerms = !!token;
+  const [selected, setSelected] = React.useState(null);
+  const [agree, setAgree] = React.useState(agreedToTerms);
+  const { data, isLoading } = useLoader(() => fetchCompendiaData(filter));
+
+  if (isLoading)
```
arielsvn: Can we use braces here? [ref](https://github.com/airbnb/javascript#blocks--braces)

src/pages/Compendia/DownloadCompendia.js
```diff
@@ -0,0 +1,112 @@
...
+    return (
+      <div className="download-compendia">
+        <p className="download-compendia__title">Fetching {title}</p>
+        <Spinner />
+      </div>
+    );
+  if (!isLoading && data.length === 0)
```
arielsvn: `!isLoading` is always true here (from lgtm)

davidsmejia: 👍 changing to an else if clause

Issue #767, opened 2019-10-03
Filter search clears unexpectedly


cgreene:
![ezgif-5-12c4aade8bee](https://user-images.githubusercontent.com/542643/66131861-26ddd880-e5c2-11e9-834f-8ce5200fda51.gif)

I expected the search to remain active within the same filter group.

arielsvn:
Fixed by #777

PR #777, opened 2019-10-17
save search filter category query in LS

davidsmejia:
## Issue Number

#767 

Since we reload the page on changes to the search query (which includes filtering) we tend to drop state for anything that isn't explicitly part of the search query. This PR preserves the state of the filter filter query so after interaction and reload of the search page the filter query will be stashed in local storage and then repopulated after page load. 

## Types of changes

## Functional tests

Works in chrome and FF

## Checklist

### DIFF
--- a/src/pages/Search/ResultFilters/FilterCategory.js
+++ b/src/pages/Search/ResultFilters/FilterCategory.js
@@ -5,6 +5,7 @@ import Checkbox from '../../../components/Checkbox';
 import Button from '../../../components/Button';
 import { InputSearch } from '../../../components/Input';
 import HighlightedText from '../../../components/HighlightedText';
+import { useLocalStorage } from '../../../common/hooks';
 
 const MAX_FILTERS = 6;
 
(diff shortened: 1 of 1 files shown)


src/pages/Search/ResultFilters/FilterCategory.js
```diff
@@ -18,12 +19,20 @@ export function FilterCategory({
   filterValues,
   activeValues,
   formatValue = x => x,
-
   onToggleFilter,
 }) {
   const [query, setQuery] = React.useState('');
+  const [savedQuery, setSavedQuery] = useLocalStorage(
```
arielsvn: Can we save this in a different place? I think we talked about refs or the store?

davidsmejia: I think we might have to use this temporarily since the filters are being regenerated dynamically. We would have to refactor the filter to persist the instance of the FilterCategory in order to keep it in a ref or state.

davidsmejia: We would have to prevent refreshing the page

arielsvn: hmm are we refreshing the whole page when filters are applied?

davidsmejia: yeah whenever the querystring changes we reload the page

PR #826, opened 2020-01-09
organism _names suffix

davidsmejia:
## Issue Number

This PR will adjust for API changes in this PR.
https://github.com/AlexsLemonade/refinebio/pull/2050


arielsvn:
Looks like we'll also need to update the downloads page here

PR #839, opened 2020-01-22
Fix missing filter category when opening search url

### LINKED ISSUE #796: Organism filters missing from search
https://www.refine.bio/search?filter_order=organism&organism=homo_sapiens

The organism filters are not displayed on that URL, even though the API returns them.

via @davidsmejia

arielsvn:
## Issue Number

close #796 

When a search URL with filters applied is visited we try to get the filters from the same query without the last filter applied, in order to get the numbers for that category.

The frontend stores the filters in an object where the data for the organism section is under `organisms` which is the same name as the parameter in the API `organisms.

In the case of the API, these fields are named differently. The organisms information is sent in a field named `organism_names` and the API parameter is `organisms`. Maybe we should consider using the same field for those? This would allow us to avoid changes like the ones here.

This PR should fix the problem while we decide.

## Types of changes

## Functional tests

Tested locally.

## Checklist

### DIFF
--- a/src/state/search/actions.js
+++ b/src/state/search/actions.js
@@ -169,7 +169,11 @@ export function fetchResults({
               [lastFilterName]: undefined,
             },
           });
-          previousFilters = previousFacets;
+          previousFilters = {
+            ...previousFacets,
+            organism: previousFacets.organism_names,
+            platform: previousFacets.platform_accession_codes,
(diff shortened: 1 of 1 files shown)


davidsmejia:
i support changing the api param to match `organism_name(s)`

Issue #841, opened 2020-01-23
Dataset request form appears to be broken.


cgreene:
@jaclyn-taroni and I attempted to request additional datasets this morning since it seemed like it had been a while since we'd received a request. We weren't able to get it to send a request but no error message was displayed.

arielsvn:
This is working now.

PR #842, opened 2020-01-23
adds getIP and handle fetch errors

davidsmejia:
## Issue Number

#841 

We need to handle errors on slack if the api endpoint for getting IP addresses fails. This adds that.

### DIFF
--- a/src/common/slack.js
+++ b/src/common/slack.js
@@ -6,6 +6,21 @@
 const SLACK_HOOK_URL =
   'aHR0cHM6Ly9ob29rcy5zbGFjay5jb20vc2VydmljZXMvVDYyR1g1UlFVL0JCUzUyVDc5OC9UWnZZSGkzcnFyY29aSWYyaFlCVDZVdTA=';
 
+// get IP
+const getIP = async () => {
+  let ip = 'N/A';
+
+  await fetch('https://api.ipify?format=json')
+    .then(resp => resp.json())
+    .then(json => json.ip)
+    .then(fetchedIP => {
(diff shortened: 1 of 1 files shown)


src/common/slack.js
```diff
@@ -6,6 +6,21 @@
 const SLACK_HOOK_URL =
   'aHR0cHM6Ly9ob29rcy5zbGFjay5jb20vc2VydmljZXMvVDYyR1g1UlFVL0JCUzUyVDc5OC9UWnZZSGkzcnFyY29aSWYyaFlCVDZVdTA=';
 
+// get IP
+const getIP = async () => {
+  let ip = 'N/A';
+
+  await fetch('https://api.ipify?format=json')
+    .then(resp => resp.json())
```
arielsvn: hmm I thought the point of await was to avoid these callbacks. Can we use one or the other? putting the `fetch` inside of a `try...catch` should have the same effect.

davidsmejia: I thought it was clearer to read than than nested awaits

src/common/slack.js
```diff
@@ -110,7 +113,7 @@ export async function submitExperimentDataRequest(accessionCode, values) {
               ]
             : []),
         ],
-        footer: `Refine.bio | ${ip} | ${navigator.userAgent}`,
+        footer: `Refine.bio | ${await getIP()} | ${navigator.userAgent}`,
```
arielsvn: I would put this in a variable to make it more explicit, but it's alright like this as well.

PR #847, opened 2020-01-29
dataset download samples fix filtering

davidsmejia:
## Issue Number

N/A

Fixes an issue where the `view samples` button was showing a list of all samples on the site... and not the ones in the dataset


src/pages/Downloads/DownloadDetails.js
```diff
@@ -103,7 +103,7 @@ ClearDatasetButton = connect(
 
 export function SpeciesSamples({
   dataSet: {
-    dataSetId,
+    id: dataSetId,
```
davidsmejia: this was where the bug was

davidsmejia:
Addressed in another PR

PR #853, opened 2020-02-11
Add server side rendering

arielsvn:
## Issue Number

#844 

## Purpose/Implementation Notes

This turned out to be more complicated than I was expecting, however, I'm happy with how it's coming along. I deployed a demo to https://refinebio.now.sh/

A few things are missing from this PR in order to be ready:

- [x] Find a solution to navigate between pages that don't rely on the URL `state` because NextJS doesn't support this by default.

- [x] Fix filters on search

- [x] Fix unit tests

- [x] Fix integration tests

- [x] Compendia page toggle

- [x] Remove all packages related to CRA

- [ ] Add deploy configuration, we talked about Lambda@Edge. Alternatively, now.sh also integrates with Github and can deploy the app automatically.

@davidsmejia feel free to take a look, I'd really appreciate any comments. With the number of changes this includes we'll have to review it carefully. 

## Types of changes

* [ ] Bugfix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Functional tests

Demo on https://refinebio.now.sh/

## Checklist

* [ ] Lint and unit tests pass locally with my changes
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] I have added necessary documentation (if appropriate)
* [ ] Any dependent changes have been merged and published in downstream modules

PR #859, opened 2020-03-04
[Deploy] Master

arielsvn:
Merge SSR changes #853


arielsvn:
That error was caused because I had added a test domain to Zeit which wasn't correctly configured. I removed it and tested [the deploy](https://zeit.co/ccdl/refinebio-frontend/llhz1l475) again.

@davidsmejia we should be able to merge this now, it's needed in order to set up both prod/staging domains in Zeit.

PR #869, opened 2020-03-12
Add project name to now config

arielsvn:
## Issue Number

N/A

Otherwise, the deploys go to a different project in ZEIT.


arielsvn:
It will make the logs of the deploys publicly accessible (example https://zeit.co/ccdl/refinebio-frontend/fjxr30vdh). Right now they are private.

Issue #862, opened 2020-03-04
Finish releasing server side rendering


arielsvn:
We discussed the roadmap for the deployment initially in https://github.com/AlexsLemonade/refinebio-frontend/pull/853#issuecomment-594640628. From those I finished today:

- [x] Merge https://github.com/AlexsLemonade/refinebio-frontend/pull/853
- [x] Create a CCDL team in Zeit, and setup both staging and prod deploys under .now.sh domains
- [x] Setup CCDL team on Zeit connected to refinebio's https://zeit.co/ccdl/refinebio-frontend

We have yet to take care of the following:

- [x] Merge https://github.com/AlexsLemonade/refinebio-frontend/pull/859 which should deploy to production
- [x] Ensure the staging/production configuration is correct for the project in Zeit. I added files to set it up with https://github.com/AlexsLemonade/refinebio-frontend/pull/861 but we still need to merge everything into master to see that it works. [Some docs](https://docs-31e8787wy.zeit.sh/docs/v2/domains-and-aliases/aliasing-a-deployment/#git-staging-aliases)
- [x] Ensure that we can have a domain pointing to the `master` branch. I setup two domains now:
![image](https://user-images.githubusercontent.com/1882507/75927446-655e1480-5e3a-11ea-92cf-5e0bf7a6139d.png)
- [x] Update refine.bio DNS configuration to point to ZEIT servers
  - [x] Staging
  - [x] Production
- [ ] Remove all terraform configuration related to the frontend from the backend project.

arielsvn:
I'm having an issue with Zeit that's very frustrating, where it's not letting me deploy a domain with the `master` branch. When I add it they deploy `dev` instead, which is the default branch of this project.

I filed a ticket with ZEIT to see if there's a way to fix this that doesn't involve changing the default branch of this repository. [Recording of the problem](https://user-images.githubusercontent.com/1882507/76360182-e4e45b80-62f2-11ea-9726-09cc39ca8110.gif)

If this doesn't get fixed soon, we could use ZEIT CLI interface with CI to have both prod and dev deployed. Similar to [this config](https://spectrum.chat/zeit/now/multiple-environments-via-github-deployments~34124374-4365-4abd-a53f-0698315adeb5?m=MTU0ODM3Njc3NTM2NQ==)

arielsvn:
[...]

PR #870, opened 2020-03-12
[master] Test production deploy with ZEIT

arielsvn:
## Issue Number

#862 

Got a staging deploy working, now I want to make sure that the production one also works.

https://refinebio-staging.now.sh/


.circleci/config.yml
```diff
@@ -19,9 +19,37 @@ jobs:
...
+
+  deploy-staging:
+    docker:
+      - image: circleci/python:3-node
+    steps:
+      - checkout
+      - run: yarn global add now
+      - run:
+          command: $(yarn global bin)/now --confirm --prod --token $ZEIT_TOKEN --local-config now.staging.json --scope ccdl
+          no_output_timeout: 1h
```
kurtwheeler: Wait the PR says production but there's both staging and prod here. Why is staging also being added?

arielsvn: We deploy both prod/staging with the same command, the only difference is the config file. Also, these are only executed for the `master`/`dev` branches

Issue #810, opened 2019-12-16
Provide a one-button mechanism for users to download all samples from a single experiment


cgreene:
Wherever we display an experiment, we could provide a button for convenience that would create a dataset and allow users to easily download all samples from a single experiment along with associated metadata.

arielsvn:
Right now it's fairly easy for users to get a dataset with a single experiment, with 4 clicks. An additional button would probably reduce it to 2 clicks. Maybe this is an indication that we should do a better job of explaining things?

![2019-12-16 10 01 56](https://user-images.githubusercontent.com/1882507/70917575-95400d80-1feb-11ea-853c-1a1f5a055ab2.gif)

dvenprasad:
We need to think about whether we provide transformation options and advance download options(QN vs non-QN) if they are downloading a single experiment.

arielsvn:
Looks like people are downloading datasets with a single experiment mostly. Out of the 95 datasets that we have processed this year, 79 had a single experiment in them. They were all Scaled by `None` and the only option that changed was whether or not they were quantile normalized (50 yes vs 29 no)

I think we should provide transformation and advance download options.

<details><summary>Code for reference</summary>

```py
>>> datasets = Dataset.objects.all().filter(created_at__gt='2020-01-01', is_processed=True)        
>>> print(datasets.count())
95
>>> count = 0
>>> ids = []
>>> for dataset in datasets:
...     if len(dataset.data.keys()) == 1:
...         count += 1
...         ids.append(dataset.id)
... 
>>> print(count)
79
>>> 
>>> Dataset.objects.filter(id__in=ids).values_list('aggregate_by', 'scale_by', 'quantile_normalize', 'quant_sf_only', 'svd_algorithm').distinct()
<QuerySet [('EXPERIMENT', 'NONE', False, False, 'NONE'), ('EXPERIMENT', 'NONE', True, False, 'NONE')]>
>>> Dataset.objects.filter(id__in=ids, quantile_normalize=True).count()
50
>>> Dataset.objects.filter(id__in=ids, quantile_normalize=False).count()
29
```

</details>

dvenprasad:
Here are the mockups:

![experiment-page-dl-now](https://user-images.githubusercontent.com/15315514/74863464-f05bec80-531b-11ea-9332-8ce6e3bafa18.png)

[...]

PR #886, opened 2020-03-19
810 download experiment

davidsmejia:
## Issue Number

#810 

Adds some helpers for moving away from redux
Adds a DownloadExperiment component
Adds a composable DownloadOptionsForm component

## Types of changes

## Functional tests

Downloaded experiment as a dataset, refreshed to see if the badge was correctly showing

## Checklist

## Screenshots

Please attach any screenshots that illustrate these changes.


src/common/helpers.js
```diff
@@ -359,3 +359,23 @@ export function numberFormatter(num, digits = 0) {
...
   return (num / si[i].value).toFixed(digits).replace(rx, '$1') + si[i].symbol;
 }
+
+/**
+ * Subscribe an email to CCDL Emails
+ *
+ * @param {*} EmailAddress
+ */
+
+export const subscribeToMailingList = async email => {
```
arielsvn: Do we use this anywhere?

src/common/helpers.js
```diff
@@ -359,3 +359,23 @@ export function numberFormatter(num, digits = 0) {
...
+  const portalId = '5187852';
+  const formGuid = 'ea5b970d-c22f-4b42-b569-26f84b1cea2f';
+  const formUrl = `https://api.hsforms.com/submissions/v3/integration/submit/${portalId}/${formGuid}`;
+  try {
+    await Ajax.post(formUrl, {
+      fields: [{ name: 'email', value: email }],
+    });
+  } catch (err) {
+    // Show error message?"
```
arielsvn: We should log the failure to somewhere, maybe sentry?

src/common/hooks.js
```diff
@@ -78,7 +78,11 @@ export function useLocalStorage(key, initialValue) {
       setStoredValue(valueToStore);
       // Save to local storage
       if (!isServer) {
-        window.localStorage.setItem(key, JSON.stringify(valueToStore));
+        if (valueToStore === null) {
```
arielsvn: Should this be `valueToStore === undefined`? Why not save `null` values?

src/components/DatasetDownloadOptionsForm.js
```diff
@@ -0,0 +1,413 @@
...
+
+// This is the mechanism for propagating properties to child components
+// TODO: Any option should be overridable at the child component level
+const DatasetDownloadOptionsContext = React.createContext();
+
+// If you do not add children this is what will be populated
+// Notice no EMAIL
+const DefaultOptions = () => {
+  return (
+    <DatasetDownloadOptionsContext.Consumer>
```
arielsvn: Is it better to use the hook version of this [useContext](https://reactjs.org/docs/hooks-reference.html#usecontext)?

src/components/DownloadExperiment.js
```diff
@@ -0,0 +1,187 @@
...
+const DownloadExperiment = ({ experiment, showProcessing = false }) => {
+  const [emailAddress] = useLocalStorage('email-address');
+
+  const {
+    accession_code: experimentAccessionCode,
+    num_downloadable_samples: downloadableSamplesCount,
+  } = experiment;
+
+  const data = {};
+  data[experimentAccessionCode] = experiment.samples.map(s => s.accession_code);
```
arielsvn: It's also possible to initialize the property directly:

```suggestion
  const data = {
    [experimentAccessionCode]: experiment.samples.map(s => s.accession_code)
  };
```

src/components/DownloadExperiment.js
```diff
@@ -0,0 +1,187 @@
...
+    };
+
+    const justStarted = dataset.id !== processingDataset;
+    const wasProccessing = processingDataset && !dataset.id;
+
+    if (datasetId && (justStarted || wasProccessing)) refreshState();
+  }, [dataset, setDataset, processingDataset, setDatasetIsProcessing]);
+
+  // Hide this component
+  if (downloadableSamplesCount === 0) return false;
```
arielsvn: Should this be the responsability of the parent? To only show the component when there're downloadable samples?

src/components/DownloadExperiment.js
```diff
@@ -0,0 +1,187 @@
...
+      if (refreshedDataset.is_processing) {
+        setDataset(refreshedDataset);
+        setDatasetIsProcessing(refreshedDataset.id);
+      } else {
+        setDataset(emptyDataset);
+        setDatasetIsProcessing(null);
+      }
+    };
+
+    const justStarted = dataset.id !== processingDataset;
```
arielsvn: in what scenario is this `false`?

arielsvn:
Also, when the "Download Now" button next to the samples table is used to download, the experiment is not marked as Processing.

src/common/helpers.js
davidsmejia: good call that should be at the end of the submit for datasetdowloadoptionsform.js

src/common/helpers.js
davidsmejia: we talked about this before... this is how it is currently implemented but im open to telling sentry.. is there any problem with passing the email or we could pass the dataset and then manually subscribe them to the mailing list if it fails, but I think it's very unlikely to fail unless maybe they have a weird internet connection drop

arielsvn: yes, sending the dataset id should be enough. Hopefully, this never fails

src/common/hooks.js
davidsmejia: Yeah im open to either one... or we could also pass back an array of 3 values with an explicit clear function so something like

```
const [token, setToken, clearToken] = useLocalStorage('token', false)
```

src/components/DatasetDownloadOptionsForm.js
davidsmejia: I like this better, good look!

src/components/DownloadExperiment.js
davidsmejia: yeah that would be a good call for this case (actually I already think it's implemented for the case when there are no downloadable samples, but I think we should still handle the case where we pass a parameter to hide when the dataset is being processed and the button shouldn't show, Just to keep the logic more dry...

src/components/DownloadExperiment.js
davidsmejia: This is false when the id's are the same, so the form just completed and updated both values to be the same.

davidsmejia: This is false when the processing dataset is the same as the local storage dataset. It will be false when the page is refreshed.

davidsmejia:
> Also, when the "Download Now" button next to the samples table is used to download, the experiment is not marked as Processing.

Yeah by design:
https://user-images.githubusercontent.com/15315514/74863463-f05bec80-531b-11ea-8e23-3f242108d65f.png

dvenprasad:
Hey! I looked at the preview and noticed some things were a little different from design, and wanted know what the rationale was.

1) Search result cards don't have `download now` buttons (maybe we talked about this? not sure) 

2) Download Now Modal:
    - Shows option to skip QN for microarray experiments. This option is only applicable to RNA-seq samples. (Can we disable it for micoarray experiments?) 
     - Shows aggregate options for experiments with only one species. (Can we disable it for single species experiment?) 

There were some aesthetic inconsistencies I noticed as well for the `Download Now Modal`
- The padding on the modal is too small. It should be 40px. 
- The spacing between label and form field is also too small. It should be 16px.

davidsmejia:
@arielsvn could you take another look at this, I was able to support the requested feature set by pulling some data management outside of the component, would love to get your thoughts on it. going to tag you for review

some things that should be working
* synchronization between navigation - download from search will show tag on experiment page
* updates on the same page - ie download form near samples section should update the top download bar too
* synchronization between tabs - download in one tab should update state on the other tab (if its the same experiment)

src/common/helpers.js
davidsmejia: I'm not really sure how we would escalate exceptions to Sentry without triggering an error boundary capture, do you have any direction for this? Also let's make it a new ticket because this isn't really a change specific to this PR it's just moving the same logic we had so it can be shared.

dvenprasad:
@arielsvn we email them the dataset. On the `We're processing` page(?)/state(?) of the modal, we provide a link to track progress, which opens up a new tab.

src/components/DatasetDownloadOptionsForm.js
```diff
@@ -0,0 +1,397 @@
...
+    >
+      <Button type="submit" text={actionText} disabled={isSubmitting} />
+    </div>
+  );
+};
+
+/**
+ *
+ */
+const DatasetDownloadOptionsFormComponent = ({
```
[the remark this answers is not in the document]
davidsmejia: Yeah this is setting up for removing redux for managing the original dataset. It's a lot of duplication but I'm not going to be including the refactor for that inside of this because it will require a lot more time to untangle that flow.

src/components/DownloadExperiment.js
```diff
@@ -0,0 +1,217 @@
...
+  EmailAddressOptions,
+  SubmitDatasetOptionsForm,
+} from './DatasetDownloadOptionsForm';
+
+import { getDataSet } from '../api/dataSet';
+
+import { useLocalStorage, useWatchedLocalStorage } from '../common/hooks';
+import { getDomain } from '../common/helpers';
+
+// TODO: Rethink how we use email fallback to support designs
```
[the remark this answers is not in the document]
davidsmejia: We don't have user accounts so we assume that the email in local store. Now we want to support multiple simultaneous datasets which potentially could be associated with multiple email addresses. The problem is that we don't receive the email from the API (for privacy reasons).

src/components/DatasetDownloadOptionsForm.js
```diff
@@ -0,0 +1,397 @@
...
+  token,
+  createToken: connectedCreateToken,
+}) => {
+  const [showAdvancedOptions, setAdvancedOptions] = React.useState(
+    !dataset.quantile_normalize || advancedOptions
+  );
+
+  const [emailAddress, setEmailAddress] = useLocalStorage('email-address', '');
+
+  const defaultValidation = () => {
```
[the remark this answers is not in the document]
davidsmejia: I think it might be possible but this current implementation requires that the token is passed down into the component which is used in the validation rules. I moved it outside and have a passable parameter now.

src/common/hooks.js
```diff
@@ -89,6 +93,71 @@ export function useLocalStorage(key, initialValue) {
   return [storedValue, setValue];
 }
 
+/**
+ * Same as useLocalStrage but updates state when local storage changes
+ */
+
+export function useWatchedLocalStorage(key, initialValue) {
```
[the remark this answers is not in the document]
davidsmejia: We might be able to replace it but that would be outside the scope of this PR I'm not sure what would break if it is expecting the existing the behavior.

src/api/dataSet.js
```diff
@@ -57,6 +64,14 @@ export async function updateDataSet(dataSetId, dataSet, details = false) {
     : result;
 }
 
+export async function updateDataSetOptions(dataset, token = false) {
```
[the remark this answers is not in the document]
davidsmejia: The one above it wasn't taking headers and had a weird return clause. I think it's better to leave this alone for now. This will be removed anyway when we move away from redux.

src/pages/experiments/index.js
```diff
@@ -106,18 +113,29 @@ let Experiment = ({
...
+                        all: true,
+                        total: numDownloadableSamples,
+                      },
+                    }}
+                  />
+                  {!dataset.is_processing ? (
+                    <DownloadExperiment
+                      experiment={experiment}
+                      dataset={dataset}
+                      setDataset={setDataset}
```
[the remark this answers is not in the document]
davidsmejia: * The experiment page wants to be able to manage the state (in order to update multiple instances of the same dataset being downloadable) 
* I think its better if we move that management outside of the component, dataset and setDataset can also be wrapped this way
* This configuration is just one, probably one of the more complex examples, but in the future other components won't be so deeply nested (ie) dataset download page with will just need to have the dataset its configuring options for and a way to update it because of the way the API works, ideally we will move away from that in the future and can treat it more restfully. 📅

PR #890, opened 2020-04-27
Don't let users try to download experiments with no samples

davidsmejia:
## Issue Number

N/A
Also updates yarn + node-sass because it was blocking


src/pages/experiments/index.js
```diff
@@ -358,7 +359,7 @@ function SamplesTableBlock({ experiment, dataset, setDataset }) {
           <div className="experiment__header">
             <h2 className="experiment__title mobile-p">Samples</h2>
             <div>
-              {experiment && (
+              {experiment && numDownloadableSamples > 0 && (
```
davidsmejia, line 362: this is the relevant change

PR #902, opened 2020-06-16
Data requests as tickets

BEW111:
## Issue Number

[#1738](https://app.zenhub.com/workspaces/refinebio-5c92300a9d1bd30b8d3101f2/issues/alexslemonade/refinebio/1738)

Experiment data requests are added as issues to GitHub, and those contacts are added to HubSpot. Right now still using test repo and HubSpot list. I couldn't find a way to change the internal name of a custom field in HubSpot but it seems that it's automatically set as `dataset_request_details` for a custom field called Dataset Request Details, so this should work with the actual contact list as long as it is named the same.

## Types of changes

GitHub token, HubSpot API key, and Slack webhook URL should be env variables.

## Functional tests

I've made a [test repo](https://github.com/BEW111/testrepo/issues) which the dataset requests are sent to, and a test list on HubSpot for new contacts. If a GitHub or HubSpot request fails then it is instead sent to Slack. As of right now I think the only way the HubSpot request would fail is if an invalid email is entered that the refine.bio form doesn't recognize as invalid but that HubSpot does deem as invalid.

## Checklist

## Screenshots

hubspottest

### DIFF
--- a/src/common/github.js
+++ b/src/common/github.js
@@ -2,7 +2,6 @@
  * This file contains helper methods that create new GitHub requests
  */
 
-// btoa
 const GITHUB_URL = 'https://api.github.com/repos/BEW111/testrepo/issues'; // this is a test github repo, change to actual repo once done
 const { GITHUB_TOKEN } = process.env;
 
--- a/src/common/github.js
+++ b/src/common/github.js
@@ -2,7 +2,8 @@
  * This file contains helper methods that create new GitHub requests
  */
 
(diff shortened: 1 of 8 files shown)


src/common/github.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * This file contains helper methods that create new GitHub requests
+ */
+
+const GITHUB_URL = 'https://api.github.com/repos/BEW111/testrepo/issues'; // this is a test github repo, change to actual repo once done
```
davidsmejia, line 5: Just leaving a comment here to make sure this gets updated. I think we should make this an env variable so we can have stage and local use an alternate repo

Issue #923, opened 2020-08-26
Sample processing modal should include Ensembl release for RNA-seq samples


jaclyn-taroni:
Context is: https://github.com/AlexsLemonade/refinebio/issues/2497#issuecomment-680907888

The `source_version` field included for a transcriptome index and returned as part of a sample's results contains information about the release of Ensembl used to build that index. The release used has an impact on the data that users download (https://github.com/AlexsLemonade/refinebio/issues/2497). 

The modal on the frontend currently only reports the genome build used. 

<img width="803" alt="Screen Shot 2020-08-26 at 8 34 17 AM" src="https://user-images.githubusercontent.com/19534205/91314655-cc72db00-e784-11ea-8d32-55c00828d6ce.png">

We should also report the release used.

PR #924, opened 2020-08-28
Include Ensembl release and division for RNA-seq samples

BEW111:
## Issue Number

#923 

This is dependent on AlexsLemonade/refinebio/pull/2527, since it uses the `release_version` and `database_name` fields of the transcriptome indices. The processing info for a transcriptome index now shows those two fields.

## Types of changes

## Functional tests

I tested this with the current API (not the changes from [#2527](https://github.com/AlexsLemonade/refinebio/pull/2527)) and `source_version` shows up, and the only difference between that and the new field names is the names themselves so it should show up as long as they are in the database.

## Checklist

## Screenshots

Right now this doesn't have the entries since those would come from the changes in [#2527](https://github.com/AlexsLemonade/refinebio/pull/2527)

Screen Shot 2020-08-31 at 12 34 26 PM

### DIFF
--- a/src/components/SamplesTable/ProcessingInformation/ProcessorVersion.js
+++ b/src/components/SamplesTable/ProcessingInformation/ProcessorVersion.js
@@ -43,6 +43,8 @@ export default class ProcessorVersion extends React.Component {
           {results.map(({ processor }) => this._renderProcessor(processor))}
 
           {this._renderGnomeVersion()}
+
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
I would probably put the new fields under `Genome Build`, rather than put that information under a new section (e.g., `Database Info`).

PR #925, opened 2020-10-01
set now@19.2.0 for deploys

davidsmejia:
## Issue Number

N/A

Deploys with github actions arent working with version 20 of now so instead of installing the latest use the most recent that complies with our current configs. ie 19.2.0 - This occurred on the resources portal and is something we should revisit in the future.

## Types of changes

## Functional tests

N/A

## Checklist


kurtwheeler (approved):
LGTM!

PR #940, opened 2021-04-06
log out stage and api host

davidsmejia:
## Issue Number

n/a log out non-sensitive deploy args to see if they are as expected


kurtwheeler (approved):
Word

Issue #926, opened 2020-10-07
Restructure examples we present on homepage 


dvenprasad:
refine.bio examples are revamped. Once it goes live, the current examples we display on the home page are not accurate anymore.

We should change how we present these to reflect the updated refinebio examples content. 

![Screen Shot 2020-10-07 at 11 12 02 AM](https://user-images.githubusercontent.com/15315514/95350871-a1ed6500-088e-11eb-9775-39ebf1afa17e.png)

Tagging @cansavvy to help figure out which examples we should link

jaclyn-taroni:
Tagging @cansavvy again

cansavvy:
I wonder whether we should instead move to linking to the intro to microarray and intro to rna-seq pages and/or the getting started page? Now that we have split things by technology I am not sure whether to link people to rna-seq differential expression or microarray differential expression. 

OR should we post both technologies examples? 

If we do go the specific example route, I would say differential expression (two groups version) for sure and probably clustering and maybe one of the pathway analyses GSEA?

dvenprasad:
We can definitely link to both technologies. I think a combination of specific examples (i.e an analysis) and something generic like getting started and intro would be good. 

Another thing, we want to provide links to things which would cover both our user types: Bio Expert and Data Whiz (ooo boy been a long time since I brought these up). I'm inclined to leave the use your data alongside refine.bio (for the Data Whiz) unless there is a more apt example/thing we can direct them to.

jaclyn-taroni:
> We can definitely link to both technologies. I think a combination of specific examples (i.e an analysis) and something generic like getting started and intro would be good.

If we use this existing structure, we now have several examples we can link to which I'll detail below.

#### Differential expression

- [Differential expression analysis: microarray - 2 groups](https://alexslemonade.github.io/refinebio-examples/02-microarray/differential-expression_microarray_01_2-groups.html)
- [Differential expression analysis: microarray - several groups](https://alexslemonade.github.io/refinebio-examples/02-microarray/differential-expression_microarray_02_several-groups.html)
- [Differential expression analysis: RNA-seq](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/differential-expression_rnaseq_01.html) (we only have a 2 group example at the moment).

#### Pathway analysis

- [Pathway analysis: Microarray - Overrepresentation Analysis](https://alexslemonade.github.io/refinebio-examples/02-microarray/pathway-analysis_microarray_01_ora.html)
[...]

Issue #906, opened 2020-07-07
Link to different examples based on the technology in user's dataset


dvenprasad:
refine.bio examples is being restructured! 🎉  It is going to be organized by technology(https://github.com/AlexsLemonade/refinebio-examples/issues/102) 

We will need to change how we link to specific examples on the dataset download page 

![Screen Shot 2020-07-07 at 11 23 54 AM](https://user-images.githubusercontent.com/15315514/86805994-6dc39780-c046-11ea-9722-eab9c8484acd.png)

We should show different links to users based on the technology of the data they downloaded. 
Will provide more specific links once the restructure is done.

cansavvy:
This should be easier to do after the restructuring in https://github.com/AlexsLemonade/refinebio-examples/pull/120 is done.

dvenprasad:
We should keep `Learn more about how we source and process our data` as is but change the other links based on the technology in user's dataset

RNA-seq links
- [Convert ENSEMBL IDs to Gene Symbols](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.html)
- [Follow an example differential expression analysis](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/differential-expression_rnaseq_01.html)
- [Follow a example pathway analysis ](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/pathway-analysis_rnaseq_02_gsea.html)
- See what other analyses you can do with refine.bio data -> link to homepage

Microarray links
- [Convert ENSEMBL IDs to Gene Symbols](https://alexslemonade.github.io/refinebio-examples/02-microarray/gene-id-annotation_microarray_01_ensembl.html)
- [Follow an example differential expression analysis](https://alexslemonade.github.io/refinebio-examples/02-microarray/differential-expression_microarray_01_2-groups.html)
- [Follow a example pathway analysis ](https://alexslemonade.github.io/refinebio-examples/02-microarray/pathway-analysis_microarray_02_gsea.html)
- See what other analyses you can do with refine.bio data -> link to homepage

Mixed Technology links
- [Convert ENSEMBL IDs to Gene Symbols for microarray data](https://alexslemonade.github.io/refinebio-examples/02-microarray/gene-id-annotation_microarray_01_ensembl.html)
[...]

PR #945, opened 2021-05-03
906 examples links

davidsmejia:
## Issue Number

#906 #926 

Adds the updates to the refine bio examples links

## Types of changes

## Functional tests

n/a looked at both types of datasets

## Checklist

## Screenshots

waiting on the tests for checking example links


jaclyn-taroni:
I think the most recent commits here are also addressing #926?

wvauclain (approved):
LGTM!

One small thing I noticed is that on small screens the two buttons get smushed next to each other, since their left and right margins are auto:
image

PR #949, opened 2021-06-07
Master Deploy

davidsmejia:
Master deploy

* sponsor button
* changes email in dataset footer and header
* upgrades to packages/deps
* fixes issue with experiment request


kurtwheeler (approved):
Oh man, first deploy since December! Let's do it!!!

Issue #950, opened 2021-06-29
I broke sample detail views by removing a parameter on the jobs endpoints


kurtwheeler:
In https://github.com/AlexsLemonade/refinebio/pull/2783 I removed the boolean parameter `nomad` because I was removing nomad. I should have made sure the frontend didn't use it before doing so. We're now getting 400s from those requests and I'm sorry!

However it's now gone and not coming back so we should stop passing it.

PR #952, opened 2021-06-29
950 removed nomad

davidsmejia:
## Issue Number

#950 

* renamed the job attributes from `nomad_job_id` to `batch_job_id`
* renamed `by_volume` to `by_queue`
* ~renames the `nomad: true` filter to `is_queued: true`~
* fixes the dashboard and executive dashboard
* reworks running jobs logic to use `is_queued`

## Types of changes

* [X] Bugfix (non-breaking change which fixes an issue)

## Functional tests

n/a

## Checklist

## Screenshots

n/a


src/api/jobs.js
```diff
@@ -16,16 +16,16 @@ export async function getRunningJobs(accessionCode) {
   const [processorJobs, downloaderJobs] = await Promise.all([
     Ajax.get(`/v1/jobs/downloader/`, {
       sample_accession_code: accessionCode,
-      nomad: true,
+      is_queued: true,
     }),
     Ajax.get(`/v1/jobs/processor/`, {
       sample_accession_code: accessionCode,
-      nomad: true,
+      is_queued: true,
```
kurtwheeler, line 23: This doesn't work. It's no longer a filter and instead is just a property that is added on every result. This is because we'd have to make potentially multiple API calls to Batch to get all the currently active job ids, then pass that to the queryset as a filter. That could make it quite slow, and I think just having that as a field on the results should be sufficient right?

src/api/jobs.js
```diff
@@ -16,16 +16,14 @@ export async function getRunningJobs(accessionCode) {
   const [processorJobs, downloaderJobs] = await Promise.all([
     Ajax.get(`/v1/jobs/downloader/`, {
       sample_accession_code: accessionCode,
-      nomad: true,
     }),
     Ajax.get(`/v1/jobs/processor/`, {
       sample_accession_code: accessionCode,
-      nomad: true,
```
kurtwheeler, line 23: This function no longer returns only running jobs, but instead all jobs with a field `is_queued` that indicates whether or not it's running.

I was about to suggest filtering this list based on `is_queued`, but I don't think the list of running jobs should be generated by making additional API call. Instead we should just look at the results of the `getDownloaderJobs()` and `getProcessorJobs()` and filter those based on `is_queued`.

src/pages/Sample/SampleDebugContext.js
```diff
@@ -69,12 +69,12 @@ export function mergeJobs(processorJobs, downloaderJobs, runningJobs) {
...
     ...job,
     type: 'processor',
-    isRunning: runningJobs.includes(job.nomad_job_id),
+    isRunning: runningJobs.includes(job.batch_job_id),
   }));
   const downloaderJobsResult = downloaderJobs.map(job => ({
     ...job,
     type: 'downloader',
-    isRunning: runningJobs.includes(job.nomad_job_id),
+    isRunning: runningJobs.includes(job.batch_job_id),
```
kurtwheeler, line 77: Why not just use the `is_queued` field that's already on the job?

davidsmejia: ok yeah that works!

src/api/jobs.js
davidsmejia: 👍

davidsmejia: probably could remove this function actually but ill leave it in there for now/ever

src/pages/Sample/SampleDebugContext.js
```diff
@@ -48,33 +47,31 @@ export async function loadData(accessionCode) {
...
     ...job,
     type: 'processor',
-    isRunning: runningJobs.includes(job.nomad_job_id),
+    isRunning: job.is_queued,
   }));
   const downloaderJobsResult = downloaderJobs.map(job => ({
     ...job,
     type: 'downloader',
-    isRunning: runningJobs.includes(job.nomad_job_id),
+    isRunning: job.is_queued,
```
kurtwheeler, line 74: It seems like you could just use `is_queued` in place of `isRunning` and not set `isRunning` at all but it's NBD.

src/api/jobs.js
kurtwheeler: Is it still used? If possible it'd be ideal to avoid a separate API call.

kurtwheeler (approved):
This LGTM, although I'd prefer if we removed `getRunningJobs()` to ensure we don't make separate API calls for that.

src/pages/Sample/SampleDebugContext.js
davidsmejia: yeah im down for that

PR #962, opened 2021-07-29
Revert "Deploy the facet changes"

kurtwheeler:
Reverts AlexsLemonade/refinebio-frontend#961

We were a little trigger happy, but it'll be fine as long as the backend deploy goes through.

I will close this if it does, but if it doesn't I don't want to be blocked by no approver.


wvauclain (approved):

kurtwheeler:
Deploy went through! We don't need this.

PR #964, opened 2021-07-30
pass dataset into the dataset processing view

davidsmejia:
# hotfix

dataset processing page cant show next step links because the dataset wasn't being passed to the component

### DIFF
--- a/src/pages/dataset/DataSetPageHeader.js
+++ b/src/pages/dataset/DataSetPageHeader.js
@@ -44,7 +44,11 @@ export default function DataSetPageHeader({ emailAddress, hasError, dataSet }) {
 
   if (is_processing) {
     return (
-      <DataSetProcessing emailAddress={emailAddress} dataSetId={dataSetId} />
+      <DataSetProcessing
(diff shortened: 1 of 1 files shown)


src/pages/dataset/DataSetPageHeader.js
```diff
@@ -44,7 +44,11 @@ export default function DataSetPageHeader({ emailAddress, hasError, dataSet }) {
 
   if (is_processing) {
     return (
-      <DataSetProcessing emailAddress={emailAddress} dataSetId={dataSetId} />
+      <DataSetProcessing
+        emailAddress={emailAddress}
+        dataSetId={dataSetId}
```
wvauclain, line 49: Is there a reason for having this prop? It doesn't look like the `DataSetProcessing` function uses it, but am I missing it being used elsewhere?

davidsmejia: good catch i was just removing it

kurtwheeler (approved):
LGTM! Will's fast

Issue #946, opened 2021-05-11
Allow unprocessed datasets to be owned by anyone


kurtwheeler:
https://github.com/AlexsLemonade/AdminItems/issues/338

As far as I can tell, the API should currently support users setting `start=True` for datasets that were created by someone else. This means that MetaSRA can create datasets, then give them to the users unprocessed, allowing the user to trigger the smasher job and download the dataset themself.

This prevents us from smashing a bunch of datasets that no one actually wants.

I think this is only a frontend change? @davidsmejia can you weigh in with details on what needs to be changed to support this?

davidsmejia:
I think the proposed change will be to add a Download Now button next to add to dataset where you can trigger a popup that provides prefilled download options (matching the existing dataset) and allows for email and terms fill in.

PR #966, opened 2021-07-30
946 unprocessed datasets download

davidsmejia:
## Issue Number

#946 

* adds a download button to dataset page
* only shown when not processed or being processed
* copies when link was shared with the share dataset component
* rewrites `DatasetLoader` and `DatasetPageHeader` as functional components
* hides regenerate options when download button is present on dataset page

Also worth mentioning these changes avoided using redux and used the components that were written for the experiment download button which is part of the effort to transition away from redux as well as modernizing some components that it uses.

* The start processing button is blue by default and not disabled but it will show an alert if the agree to terms are not selected.

## Types of changes

## Functional tests

n/a tested downloading

## Checklist

* [X] I have added necessary documentation (if appropriate)

### processed dataset page with no extra button
Screen Shot 2021-07-30 at 5 03 01 PM

### share page with clone dataset button "Download Dataset"
Screen Shot 2021-07-30 at 5 02 52 PM

### dataset options are preselected from shared dataset
Screen Shot 2021-07-30 at 5 17 28 PM

### DIFF
--- a/src/pages/dataset/index.js
+++ b/src/pages/dataset/index.js
@@ -60,6 +60,7 @@ export default function DataSet() {
       {({ dataSet, isLoading, hasError: requestHasError }) => {
         if (isLoading) return <Spinner />;
         if (requestHasError) return <NoMatch />;
+
         const showDownloadButton = !(
           dataSet.is_processing || dataSet.is_processed
         );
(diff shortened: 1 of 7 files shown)


src/pages/dataset/index.js
```diff
@@ -66,6 +94,79 @@ export default function DataSet() {
...
+      )}
+      modalProps={{ center: true, className: 'modify-dataset-modal' }}
+    >
+      {({ hideModal }) => (
+        <>
+          <DatasetDownloadOptionsForm
+            dataset={dataSet}
+            setDataset={newDataset => {
+              hideModal();
+              window.location = `/dataset/${newDataset.id}`;
```
davidsmejia, line 122: Update this to use `download_url` after approval

davidsmejia:
notes:
* hide regenerate form on share view
* move download to the right hand side

dvenprasad (approved):
LGTM
🦕 🦕 🦕

src/pages/dataset/index.js
davidsmejia: this should stay as is

kurtwheeler (approved):
LGTM! Nice work!

PR #1008, opened 2023-01-11
ran npm audit

davidsmejia:
## Issue Number

N/A

some packages are going to be broken, luckily they are mainly tied to the charts we use for the dashboard, I am going to temporarily remove them until we have the new site up

- updated all packages based on npm audit
- global: changed Link to use legacy behavior to allow for nested anchor tags
- global: changed img -> Image
- global: import Image when used
- removed `withImages` from next/config
- misc changes to dep versions for compatibility

Steps:

```bash
$ rm yarn.lock
$ npm audit fix --force
$ yarn import
$ rm package-lock.json
```


nozomione (approved):
LGTM

